PR #243, opened 2015-03-18
Initial work at using both old, and new, chocolatey

gep13:
This is an interim step, calling directly into Choco.exe, while we wait for the Chocolatey.dll to a place where we can use it.

I haven't implemented everything yet, but have at least got ChocolateyGUI loading, reading in the list of installed packages, based on a direct call to choco.exe.


RichiCoder1:
Nice!

gep13:
@RichiCoder1 any chance you can have a look over this, and let me know what you think.

It is still very much a work in progress, so don't merge this :-)

Just looking for some feedback about the splitting up of the ChocolateyService in to PowerShellChocolateyService and CSharpChocolateyService.  The PowerShellChocolateyService is pretty much exactly the same as the old ChocolateyService, but it allows extension of the necessary methods to call either into old chocolatey, or new chocolatey based on what is discovered.

gep13:
@RichiCoder1 you beat me to it.  Any thoughts/concerns, would be much appreciated.  I am signing off for the night, but will pick up any feedback in the morning :smile:

RichiCoder1:
Will do! I'll try and give it a rough review tonight. Your approach is largely the same way I was attempting to solve it.

gep13:
>@RichiCoder1 said...
Will do! I'll try and give it a rough review tonight. Your approach is largely the same way I was attempting to solve it.

Cool, glad we are on the same page :-)  Obviously, we can replace the interim Process.Start's with direct calls to the DLL when it becomes available.

I don't "love" what I had to do with the AutoFac Registration, feels a bit clunky, so if you have any thoughts in this area, would like to hear them.

Right, i'm off, talk to you later.

RichiCoder1:
The first 3 or 4 commits could probably can and should -be cherry-pick'ed(?) into their own PR and merged.

Source/ChocolateyGui/App.xaml.cs
```diff
@@ -21,7 +22,20 @@ public partial class App
     {
         static App()
         {
-            Container = AutoFacConfiguration.RegisterAutoFac();
+            Container = AutoFacConfiguration.InitialRegistration();
+
+            var configurationProvider = Container.Resolve<IChocolateyConfigurationProvider>();
```
RichiCoder1: I feel like this should be inside the AutoFac registration logic, rather than having it in App(). No Initial/Remaining, just one intelligent registration. Especially since it doesn't make much sense to do a Resolve when registration knows exactly what to expect.

RichiCoder1:
Looks good so far besides my two comments.

gep13:
>@RichiCoder1 said...
The first 3 or 4 commits could probably can and should -be cherry-pick'ed(?) into their own PR and merged.

I actually already have them merged locally on the develop branch, just not on origin.  I will fix that up later.

Source/ChocolateyGui/App.xaml.cs
gep13: So, are you saying just to "new" up an instance on IChocolateyConfigurationProvider, within the AutoFacConfiguration class, and then do the remaining registration?  I need an instance of that class, in order to know what version of Chocolatey is being used.

gep13:
There are a few more changes that I need to make to the new interface definition, but I will keep working at it, and push new commits here when they are ready for review.  It likely won't be tonight, but I have some time over the next couple nights, so hoping to get the initial version done before the end of the weekend, but will see what I get done.

Source/ChocolateyGui/App.xaml.cs
RichiCoder1: That's exactly what I'm saying. Something like:
https://gist.github.com/RichiCoder1/a1e6c049777aca7c2093
(Note: haven't actually compiled that)

gep13: Cool, makes sense to me :+1:

PR #247, opened 2015-03-24
Ability to Export currently installed packages

gep13:
- Using a simple PersistenceService, i.e. Open/Save Dialog
- Provide option to export the current list of installed packages to a packages.config file

### DIFF
--- a/Source/ChocolateyGui/ViewModels/Controls/LocalSourceControlViewModel.cs
+++ b/Source/ChocolateyGui/ViewModels/Controls/LocalSourceControlViewModel.cs
@@ -11,8 +11,10 @@ namespace ChocolateyGui.ViewModels.Controls
     using System.Collections.ObjectModel;
     using System.ComponentModel;
     using System.Globalization;
+    using System.IO;
     using System.Linq;
     using System.Reactive.Linq;
+    using System.Text;
     using System.Threading.Tasks;
     using System.Windows;
     using ChocolateyGui.Base;
@@ -27,7 +29,9 @@ namespace ChocolateyGui.ViewModels.Controls
         private readonly ILogService _logService;
         private readonly List<IPackageViewModel> _packages;
         private readonly IProgressService _progressService;
+        private readonly IPersistenceService _persistenceService;
         private bool _hasLoaded;
+        private bool _exportAll = true;
         private bool _matchWord;
         private bool _showOnlyPackagesWithUpdate;
         private ObservableCollection<IPackageViewModel> _packageViewModels;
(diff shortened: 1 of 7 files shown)


gep13:
@RichiCoder1 can I get you to case your eyes over this? Thanks!

Source/ChocolateyGui/ViewModels/Controls/LocalSourceControlViewModel.cs
```diff
@@ -183,6 +197,37 @@ public bool ReceiveWeakEvent(Type managerType, object sender, EventArgs e)
...
+            try
+            {
+                var fileStream = this._persistenceService.SaveFile("*.config", "Config Files (.config)|*.config");
+
+                if (fileStream == null)
+                {
+                    return;
+                }
+
+                using (StreamWriter sw = new StreamWriter(fileStream, Encoding.ASCII))
```
RichiCoder1: Call me stickler, but I think we should be using actual Xml writing tools here.

Source/ChocolateyGui/ViewModels/Controls/LocalSourceControlViewModel.cs
```diff
@@ -108,6 +117,11 @@ public bool CanUpdateAll()
             return this.Packages.Any(p => p.CanUpdate);
         }
 
+        public bool CanExportAll()
+        {
+            return this._exportAll;
```
RichiCoder1: If this is never false, you should be able to just leave of this from the command. Though, thinking about it now, we might want to set _exportAll = false inside ExportAll() so they don't accidentally try and export twice.

gep13: Yeah, that is probably a good shout :+1:

Source/ChocolateyGui/ViewModels/Controls/LocalSourceControlViewModel.cs
gep13: Yip, this is just a PoC, to see what people's thoughts are.  Can switch to using Xml based classes for actual implementation.

gep13:
@RichiCoder1 okay, made some changes based on your feedback, let me know what you think

RichiCoder1:
Looks good.

gep13:
What are your thoughts on something like this:

http://www.c-sharpcorner.com/UploadFile/mahesh/wpf-dropdownbutton-control/

I thought the export button could be overridable, i.e. default would be to create a packages.config file, but it could be selected to export to boxstarter script, or boxstarter url, or chocolatey package.

This would be a future enhancement, not for now, but just something to think about.

RichiCoder1:
I'd be cool with that.

gep13:
:+1:

Issue #253, opened 2015-03-27
Use the global config sources


bc3tech:
I've added a source to my chocolatey sources, but when I launch the GUI it doesn't use those sources and I have to add it, yet again, to the app.

The app would be doubly cool if it used the choco global sources - either solely or in conjunction with its own set of sources.

gep13:
@bc3tech thanks for the suggestion!

In all honesty, I thought that this was the way that ChocolateyGUI already worked.  Let me investigate, and I will see what can be done.

RichiCoder1:
Likely done actually.

PR #257, opened 2015-03-30
(GH-253) Include chocolatey's global sources in the source list

bc3tech:
Takes chocolatey's global sources list, unions with the app-specific source list. Filters out any name|uri matches.

Also modified the determination of Choco Install location & Executable usage to be `Lazy<T>`

Note: This is an interface change to `IChocolateyConfigurationProvider`

### DIFF
--- a/Source/ChocolateyGui/Providers/ChocolateyConfigurationProvider.cs
+++ b/Source/ChocolateyGui/Providers/ChocolateyConfigurationProvider.cs
@@ -7,65 +7,80 @@
 namespace ChocolateyGui.Providers
 {
     using System;
+    using System.Collections.Generic;
     using System.IO;
     using System.Linq;
+    using System.Xml.Linq;
 
     public class ChocolateyConfigurationProvider : IChocolateyConfigurationProvider
     {
-        private string _chocolateyInstallLocation;
-        private bool _isChocolateyExecutableBeingUsed;
+        private readonly Lazy<string> _chocolateyInstallLocation;
+        private readonly Lazy<bool> _isChocolateyExecutableBeingUsed;
 
         public ChocolateyConfigurationProvider()
         {
-            this.GetChocolateyInstallLocation();
-            this.DetermineIfChocolateyExecutableIsBeingUsed();
+            this._chocolateyInstallLocation = new Lazy<string>(this.GetChocolateyInstallLocation);
+            this._isChocolateyExecutableBeingUsed = new Lazy<bool>(this.DetermineIfChocolateyExecutableIsBeingUsed);
         }
 
         public string ChocolateyInstall
         {
-            get
-            {
-                return this._chocolateyInstallLocation;
-            }
+            get { return this._chocolateyInstallLocation.Value; }
         }
 
         public bool IsChocolateyExecutableBeingUsed
         {
-            get
-            {
-                return this._isChocolateyExecutableBeingUsed;
-            }
+            get { return this._isChocolateyExecutableBeingUsed.Value; }
         }
 
-        private void GetChocolateyInstallLocation()
+        public IEnumerable<Tuple<string, string>> Sources
         {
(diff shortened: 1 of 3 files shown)


RichiCoder1:
This is more a general question, but is there any advantage to making the install location and exe calls `Lazy<T>`? If I recall correctly, they're used almost immediately after the Provider's instantiation.

Source/ChocolateyGui/Providers/ChocolateyConfigurationProvider.cs
```diff
@@ -7,65 +7,80 @@
...
-            var exePath = Path.Combine(this._chocolateyInstallLocation, "choco.exe");
+            var exePath = Path.Combine(this._chocolateyInstallLocation.Value, "choco.exe");
+
+            return File.Exists(exePath);
+        }
 
-            if (File.Exists(exePath))
+        private IEnumerable<Tuple<string, string>> GetChocolateySources()
+        {
+            var chocoConfigFile = Path.Combine(this._chocolateyInstallLocation.Value, "config", "chocolatey.config");
```
RichiCoder1: More of a note to self than anything, but this should eventually use Choco's API.

bc3tech: Completely agree. I didn't see such a capability, or know of one. Do point me in that direction if there is one - would love to play with it (and use it here of course) :)

bc3tech:
@RichiCoder1 When i saw the values being cached it screamed "Lazy<T>!" at me, so I did it. Happy to remove if it's not approved of.

Not sure what the appveyor build failure would be here.....

Source/ChocolateyGui/Providers/ChocolateyConfigurationProvider.cs
RichiCoder1: I believe @jaykul was working on some PR's for choco to allow enumeration of commands, but I don't believe it's merged (though I haven't double checked recently). Once that's in, that's what we'd use.

RichiCoder1:
I think making them lazy is probably unneccessary here.

Other wise lgtm.

bc3tech:
@RichiCoder1 is that to say you want me to revert? Wait for other opinions?

gep13:
>@bc3tech said...
Working on the AppVeyor failure due to CodeAnalysis. For some reason that didn't pop in my IDE...

Make sure and do a release build within Visual Studio.

Also, at the command line, you can do:

```.\build.bat InspectCodeForProblems Release```

to do a complete analysis, prior to pushing a Pull Request.

gep13:
>@bc3tech said...
I also wasn't entirely sure whether I should've branched from develop or master for this particular repo... i chose develop

Yes, you chose the right branch.  We are flowing gitflow, as described here:

http://nvie.com/posts/a-successful-git-branching-model/

Only hotfixes and new releases should be merged into master.  In your case, this suggestion would be part of a new release.

gep13:
I need to have a proper look at this, as I had actually started spiking an implementation, which takes a different path than you seem to have.  I only have this locally just now, but I will tidy up a few things after I have had something to eat, and put it somewhere so that we can all review it.

gep13:
@bc3tech it is difficult to see what changed in the last file in the diff, as it looks like all lines were affected.  Was this intentional?

bc3tech:
@gep13 i noticed that too.. not sure if it was VS going Castro on CR LF or tabs v spaces....

bc3tech:
wow. just noticed this in my pull reqs area... any reason we haven't merged, @gep13 ?

gep13:
@bc3tech simply put, lack of time, and other commitments :cry:

This is still on my radar, and I am hoping to circle back to ChocolateyGUI soon, but I am working on a few other things just now.

RichiCoder1:
Time commitments, and this should be technically obsoleted at this point by the chocolatey integration work happening elsewhere.

bc3tech:
fair 'nuf.

PR #258, opened 2015-03-30
(GH-253) Initial working

gep13:


gep13:
@RichiCoder1 and @bc3tech can you take a look at this?

This is far more complete, so don't be too harsh :smile_cat: 

The route that I took was to create a new implementation of ISourceService which would be used specifically for the C# version of Chocolatey, i.e. note the IoC registration.

That way, the existing functionality stays as it is, for people still using old Chocolatey, but those on new Chocolatey can start to use it, and we can trust on the API (or for now, the choco.exe commands) to work with the sources, without having to manually read and edit the config file.

What are your thoughts on this approach?

I see this solution be the more maintainable, as we rely on Chocolatey working with the config file, not us.  Also, it makes things like adding and removing the sources easier, as we call Chocolatey to do that work.

This approach completely does away with the Settings.Default.Sources version of Sources list, and instead reading solely from the config file.

Let me know what you think.

Source/ChocolateyGui/AsyncProcess/ProcessEx.cs
```diff
@@ -17,6 +17,7 @@ namespace ChocolateyGui.AsyncProcess
     using System.Diagnostics;
     using System.Threading;
     using System.Threading.Tasks;
+    using System.Windows.Documents;
```
RichiCoder1: What requires this?

Source/ChocolateyGui/Services/ChocoSettingsSourceService.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                sourcesChangedEvent(
+                    this,
+                    new SourcesChangedEventArgs(new List<SourceViewModel> { sourceViewModel }, new List<SourceViewModel>()));
+            }
+        }
+
+        public SourceViewModel GetDefaultSource()
+        {
+            throw new System.NotImplementedException();
```
RichiCoder1: Does this not cause issues elsewhere?

Source/ChocolateyGui/Services/ChocoSettingsSourceService.cs
```diff
@@ -0,0 +1,87 @@
...
+                    sources.Add(new SourceViewModel { Name = parts[0].Trim(), Url = parts[1].Trim() });
+                }
+            }
+
+            return sources;
+        }
+
+        public void RemoveSource(SourceViewModel viewModel)
+        {
+            throw new System.NotImplementedException();
```
RichiCoder1: If we're going to provide an add, shouldn't we provide a remove too?

Source/ChocolateyGui/AsyncProcess/ProcessEx.cs
gep13: Ha ha, good catch.  Nothing, no idea how that got in there.  Rebased and pushed.

RichiCoder1:
See comments.
If we do go ahead with this, we'll need to ensure that the sources in the GUI config are migrated over for the few that used them that way.

Source/ChocolateyGui/Services/ChocoSettingsSourceService.cs
gep13: In what way?  I have done initial testing, and all seems well so far, i.e. sources are read, and tabs created.  Haven't tested everything though.

RichiCoder1: The interface requires this for a reason, so I assume something calls it. And if something calls it, it's going to cause errors somewhere. If it's just simply not used, we should get rid of this on the interface.

Source/ChocolateyGui/Services/ChocoSettingsSourceService.cs
gep13: Yes, we will.  As I mentioned, this is not a complete implementation.  Just putting it out there for conversation.  This was a quick spike that I did the other day, and I was mulling over it, when I saw what Brandon was working on, I thought I better do something with it.

What I have here is just enough to get the application to open :smile_cat:

gep13:
>@RichiCoder1 said...
If we do go ahead with this, we'll need to ensure that the sources in the GUI config are migrated over for the few that used them that way.

That is a very good point!

Source/ChocolateyGui/Services/ChocoSettingsSourceService.cs
gep13: Ah sorry, I see what you are saying, I thought you were referring to something else.  See other comment about this not being a complete implementation, only for conversation.

If we want to go down this route, I would complete.

bc3tech: perhaps putting a message in the ctor of the exception? or a // TODO above? Just so we don't forget :)

bc3tech:
I'm curious if the `Process` stuff is really necessary... Isn't choco coded in c# such that we could reference the libs? I haven't dug too deep in to the actual implementation of how it works, etc.

I'm curious if/how this stuff would work if somebody ran choco from powershell. from git bash, from cmd, etc...

Not that I discount the intention, just seems that running processes to get their output is a bit "hacky" to me.

RichiCoder1:
It's incredibly hacky fix, and why I badgered the chocolatey maintainers for the longest time for a C# library. It's still a work in progress though, so many of the things we need for the gui are still missing.

gep13:
>@bc3tech said...
I'm curious if the Process stuff is really necessary... Isn't choco coded in c# such that we could reference the libs? I haven't dug too deep in to the actual implementation of how it works, etc.

Yes, you are absolutely right.

Where ChocolateyGUI is right now is really an interim state.  We want to continue to support the PowerShell version of Chocolatey, but we also want to support the C# version, now that it has been released.  Although there is a Chocolatey.dll available for consumption, it really isn't in a state where it can be put to use.  The guys who are creating the Chocolatey One-Get Provider are in the process of updating the Chocolatey.dll to make it usable in these type of scenarios.  Once the Chocolatey.dll is ready, you will see all the ```Process``` steps go away, in favour of calling into the DLL.

Hope that makes sense.  If not, feel free to ask more questions.

bc3tech:
Would it be worth parsing `chocolatey.config` directly, then, in the interim? I'm a bit wary as to how spawning processes will work w/ users on machines that might be more locked down, have corporate "spyware" installed, etc.

gep13:
I would really rather about direct parsing of the config file.  This work is already being done by Chocolatey, so I don't see the point in duplicating that effort.  If spawning processes is going to be an issue, then we have a much larger problem, as that is how all the install/uninstall is currently working for ChocolateyGUI when using Chocolatey 0.9.9.x.

bc3tech:
roger that. I suppose what we're saying is make the best of a bad situation then :)
seems sound enough, once the `NotImplementedException` spots are taken care of ;)

gep13:
Yip, that pretty much sums it up! 

Sounds like we are all on the same page then :smile_cat: 

I will try to complete the implementation tonight, and send it out for review.

bc3tech:
Any way we could put `// TODO` or `// HACK` statements in so we don't forget to tweak it when Chocolatey.dll gets finished?

gep13:
@RichiCoder1 @bc3tech ...

okay, I have completed the first pass at the full implementation of adding/removing/getting sources, using Choco vNext.

Can you let me know what you think?

@bc3tech assuming that the build completes, can you grab the msi from the AppVeyor artifacts tab and give it a run on your machine?

I have tested it locally, and seems to be working for me, but wanted some further validation.  Thanks!

Source/ChocolateyGui/Services/ChocoSettingsSourceService.cs
```diff
@@ -0,0 +1,118 @@
...
+            this.RaiseSourceAddedEvent(sourceViewModel);
+        }
+
+        public SourceViewModel GetDefaultSource()
+        {
+            var defaultSourceName = Settings.Default.currentSource;
+            var defaultSource =
+                this._sources
+                    .FirstOrDefault(
+                        s => string.Compare(defaultSourceName, s.Name, StringComparison.OrdinalIgnoreCase) == 0);
```
bc3tech: curious if you know something i don't. is `string.Compare(` better/faster than doing `defaultSourceName.Equals(`? I prefer the latter for quick readability :)

Source/ChocolateyGui/Services/ChocoSettingsSourceService.cs
```diff
@@ -0,0 +1,118 @@
...
+                return this._sources;
+            }
+
+            var arguments = new StringBuilder();
+            arguments.AppendFormat(CultureInfo.InvariantCulture, "source list");
+
+            var results = ProcessEx.Run(this.chocoExePath, arguments.ToString());
+
+            this._sources = (from source in results.StandardOutput 
+                    select source.Split('-')
```
bc3tech: since URLs can contain `-`, i think `source.split(new[] { " - " })` would be less prone to errors

Source/ChocolateyGui/Services/ChocoSettingsSourceService.cs
gep13: In all honesty, I am not sure :smile_cat:

Source/ChocolateyGui/Services/ChocoSettingsSourceService.cs
gep13: Yip, good point, I can fix that up.

Source/ChocolateyGui/Services/SettingsSourceService.cs
```diff
@@ -63,13 +54,7 @@ public void RemoveSource(SourceViewModel viewModel)
...
 
             Settings.Default.sources.Remove(string.Format(CultureInfo.CurrentCulture, "{0}|{1}", viewModel.Name, viewModel.Url));
-            var sourcesChangedEvent = this.SourcesChanged;
-            if (sourcesChangedEvent != null)
-            {
-                sourcesChangedEvent(
-                    this,
-                    new SourcesChangedEventArgs(new List<SourceViewModel>(), new List<SourceViewModel> { viewModel }));
-            }
+            this.RaiseSourceRemovedEvent(viewModel);
```
bc3tech: Seeing these makes me wonder if `Settings.Default.sources` could be made in to an `ObservableCollection<T>` to get the `Changed` events automagically? not sure how well the Settings stuff takes to having other collection types, but just a thought to get .Net to do some of the lifting?

bc3tech:
@gep13 I don't see an "AppVeyor artifacts tab" on my screen anywhere.... #noob

gep13:
>@bc3tech said...
I don't see an "AppVeyor artifacts tab" on my screen anywhere.... #noob

We use AppVeyor (http://www.appveyor.com/) to run the CI builds for ChocolateyGUI.  We build all Pull Requests, as well as checkins to develop and master branch.

When a Pull Request branch is built, there is no deployment to anywhere, there are just the build artifacts within AppVeyor itself.

When the develop branch is built, the output of the build is published to our Develop MyGet Feed here:https://www.myget.org/F/chocolateygui/

When the master branch is built (and there is no tag associated with it) the output of the build is deployed to our Master MyGet Feed here: https://www.myget.org/F/chocolateygui_master/

When the master branch is built (and it includes a tag) the output of the build is deployed to Chocolatey.org.

You can see all the AppVeyor builds here:

https://ci.appveyor.com/project/GaryEwanPark/chocolateygui

So for this pull request, you would have a look here:

https://ci.appveyor.com/project/GaryEwanPark/chocolateygui/build/0.14.0-PullRequest.258+9%20(Build%20326)

and then on the artifacts tab, you should see both a ChocolateyGUI.msi as well as a nupkg package.  You can download and install either one of these, and it should have the new Sources work that is included in this Pull Request.

Let me know if you run into any problems with it.

Thanks!

Source/ChocolateyGui/Services/SettingsSourceService.cs
gep13: In all honesty, I am not sure.  I am hesitant to do any work on the Settings.Default.soures, as I see this as essentially legacy code, that will only be used by people that are still using ChocolateyGUI in combination with the PowerShell version of Chocolatey.  I think we should encourage people to upgrade to the latest Chocolatey, and eventually we can start removing some of the legacy code.

I would be happy to accept a Pull Request to make changes to the Settings.Default.sources, but I don't think I will be making any changes in this area.

bc3tech:
ah, perfect. I thought I was expected to have an "AppVeyor" tab somewhere on my screen here in github :)

gep13:
Ah, nope, nothing quite as fancy as that.  You can get to the build from this page directly by clicking on AppVeyor Details link in the "All is well" section:

image

bc3tech:
Worked mostly well except for one weird thing.

I had never installed it on this machine, so I did so right from AppVeyor.
First launch, I saw `chocolatey` and `Chocolatey` (hence my implementation doing the case-insensitive union on the names ;) )

Then I shut it down, went to choco on the command line and added a new source.
Launched the app and now saw **only** `chocolatey` and `testserver` - notice the absence of `Chocolatey`

I think if you solve that, we're good.

gep13:
Thanks for trying this out.

Hmm, yes, very weird, I will dig into that, and let you know what comes up.  I am away to dedicate some time to chocolatey.org over the next few days, so I might not circle back around onto this until next week.  Will keep you posted though.

RichiCoder1:
Is this PR still relevant @gep13?

RichiCoder1:
I'm going to go ahead and close this out. Let me know if you'd like to pick this up again.

gep13:
@RichiCoder1 given the refactoring that you have done, I don't think this is relevant anymore, no.  One thing that I would say though, is...

When we ship the release of 0.14.0, if there are any custom sources defined in the ChocolateyGUI configuration file, should we go ahead and add them into Chocolatey properly?

RichiCoder1:
Chocolatey proper.

gep13:
>@RichiCoder1 said...
Chocolatey proper.

Sorry, not sure what you meant here.

RichiCoder1:
Sorry haha. I mean that chocolateyGUI's sources are chocolatey's sources now. No customer sources :)

gep13:
Exactly, but if there were some created in the custom source file in a previous version of the application, I think the new version of ChocolateyGUI should take responsibility for moving those custom sources to the actual Chocolatey source.

RichiCoder1:
Hmm. I hand't considered that. So much to do >_<

gep13:
@RichiCoder1 ha ha, I know that feeling.  It was just something that occurred to me, that would be quite jarring for people who had already created sources.  It wouldn't be a big deal for them to add them again, but I think that is something we would want to avoid, if possible.

PR #259, opened 2015-04-01
Hotfix/0.13.2

gep13:


gep13:
@RichiCoder1 what are your thoughts on the following:

image

gep13:
@ScottIsAFool also, what are your thoughts on the above?

ScottIsAFool:
Why not provide a "would you like to update chocolatey now?" option to that? Why make the user go through another step (a step not actually detailed in that message).

gep13:
>@ScottIsAFool said...
Why not provide a "would you like to update chocolatey now?" option to that? Why make the user go through another step (a step not actually detailed in that message).

I had a feeling you were going to say that :smile_cat:

ScottIsAFool:
Then you should have pre-empted me :P

gep13:
Yeah, well, I was being lazy :smile_cat:

RichiCoder1:
Looks good to me :+1:

gep13:
Re-implemented directly in develop branch, since this PR is outdated.

Issue #265, opened 2015-06-05
Cannot search for packages AND only show those with updates


twpol:
If I enter a search query _and_ select "Show Only Packages with Updates", I simply get the packages matching the search query. No filtering by available updates is performed. This would simply make sense to do given the UI.

gep13:
@twpol thanks for filing this.  This is a known limitation of the application at the minute.

RichiCoder1:
I'm pretty sure I resolved this issue.

PR #271, opened 2015-07-25
GH-2665 Allowed you to search based on both only those with updates a…

Shiney:
…nd query

Previously when there was a query and the Search for only those with
updates box was checked the packages were only filtered by the search
query now they are filtered by both,

This fixes issue https://github.com/chocolatey/ChocolateyGUI/issues/265


gep13:
@Shiney was there a reason that you closed this?

Shiney:
I need to talk to my employer about the CLA

gep13:
Ah, gotcha :+1:

Issue #267, opened 2015-06-29
Intergrate with Chocolatey API


RichiCoder1:
See:
https://github.com/chocolatey/choco/pull/225

Still in flight PRs:
https://github.com/chocolatey/choco/pull/226

Tasks:
* [x] Install
* [x] Uninstall
* [x] Update
* [ ] Get Latest
* [x] List
* [x] Search
* [ ] Sort List
* [x] Chocolatey-powered sources

RichiCoder1:
Working in progressing PR here: #282

RichiCoder1:
Updated PR. Updated issue with task list.

gep13:
> Updated issue with task list.

Wonder where you got that idea! :stuck_out_tongue:

RichiCoder1:
Done. Polishing now.

PR #282, opened 2015-09-24
(GH-267) Using chocolatey.lib for local operations.

RichiCoder1:
ChocolateyGUI is now using chocolatey.lib for all "local" operations. Namely list -lo, install, uninstall, upgrade, and sources. Only "remote" operations are still using the OData provider.


gep13:
@RichiCoder1 Just been looking through this, and it looks good to me!  I am loving the amount of code that we can now remove :+1: 

I have corrected a few Code Analysis, InspectCode errors, etc, but there are still some left.  They are regarding the `Grid.Row` and `Grid.Column` values for parts of the XAML.  Would you be able to correct these issues?  The issues are:

image

Thanks!  There is still an issue with the build that I will need to look into tomorrow, but once that and the InspectCode issues are corrected, we should be in a position to merge this into the develop branch! :smile:

gep13:
@RichiCoder1 ok, I see what the issue with the build is, I will fix that up tomorrow morning.  Cheers!

gep13:
Oh, that's weird!  When running on AppVeyor, the errors that I am seeing locally are only warnings:

image

Not sure why they are reacting differently, as they should be using the same configuration file.

Anyway, if you are happy, I can merge this into the develop branch, and look to start a new release.  Thoughts?

RichiCoder1:
I'd be ok with merging, but I don't want to release this until we've got a sufficient level of polish around things like progress messages and command execution dialogs.

PR #289, opened 2015-10-12
Use pre-release CLI tools.

RichiCoder1:


RichiCoder1:
For some reason, Dupfinder throws the warning (and therefore fails AppVeyor): "Warning: Could not load the catalog assemblies. Reference from assembly JetBrains.ReSharper.UnitTestRunner.nUnit30 to assembly nunit.core.engine could not be resolved."

But it only throws in on *first run*. I'm baffled.

RichiCoder1:
Resharper tracking issue: https://youtrack.jetbrains.com/issue/RSRP-446024

Issue #331, opened 2016-09-20
missing packages in ChocolateyGUI


mwallner:
ChocolateyGUI doesn't show all packages on source-feed (simple ChocolateyServer) and pagination doesn't start when needed. (pagination starts when I select 'All Versions' - but there are still packages missing)
There are always packages missing in ChocolateyGUI that are visible via commandline 'choco list'.
(currently 104 packages)
- sometimes the missing packages can be found via the search function (after hitting 'refresh packages' a couple of times)

mwallner:
I guess the issue is in 
ChocolateyGui\ViewModels\Controls\RemoteSourceControlViewModel.cs

- the PageCount should probably be calculated like this:
`this.PageCount = (int)(((double)result.TotalCount / (double)this.PageSize) + 0.5);`

PR #335, opened 2016-11-11
(GH-331) missing packages in ChocolateyGUI

mwallner:
The calculation of needed pages (PageCount) was wrong because of an invalid integer calculation where a floating-point operation would be needed.

Without this change there might be not enought pages to display all packages.
Example: 60 packages in repo - only 1 page containing 50 packages would be visible


RichiCoder1:
Looks like this is still relevant. @mwallner, could you bring this update to date w/ the latest develop?

mwallner:
@RichiCoder1 yup, first thing tomorrow.

Issue #337, opened 2016-11-14
FauxPowerShellConsole: OrangeRed as 'Warning' very misleading


mwallner:
OrangeRed looks very similar to Red - which suggests an error occured.
Given that chocolatey (choco.exe) itself prints 'warnings' to report how many packages were installed etc. on each run this should be changed to something different. (Yellow?)

RichiCoder1:
Closed w/ c4210da

PR #338, opened 2016-11-14
(GH-337) FauxPowerShellConsole: OrangeRed as 'Warning' very misleading

mwallner:
switched warning color from OrangeRed to Yellow


RichiCoder1:
@mwallner Thank you for this! When I original wrote this, I was under the (wrong) impression that the color would be more orange, and less red. Merged this change into feature/Framework_Overhaul c4210da045

PR #345, opened 2016-12-09
(GH-331) missing packages in ChocolateyGUI

mwallner:
The calculation of needed pages (PageCount) was wrong because of an invalid integer calculation where a floating-point operation would be needed.

Without this change there might be not enought pages to display all packages.
Example: 60 packages in repo - only 1 page containing 50 packages would be visible


A-AontwSR:
The calculation of PageCount is still not correct.
Example, with the current floating-point calculation:
`PageCount = (int)(((double)result.TotalCount / (double)PageSize) + 0.5);`
if you have 60 packages and 50 items per page, you only get 1 page, the last 10 items are missing, because:
`PageCount = (int)((60.0 / 50.0) + 0.5) = (int)(1.2 + 0.5) = (int)1.7 = 1`

You should use the Ceiling() function to round up to the correct PageCount:
```c#
PageCount = Math.Ceiling((double)result.TotalCount / (double)PageSize);
```

PR #346, opened 2016-12-09
Add UAC Prompt & Configuration

RichiCoder1:
This PR does a number of things.

- Changes the default ChocolateyGUI elevation to user.
- Adds a first run configuration to allow always starting as administrator
-Adds UAC shields to signify "privileged" commands when running as a non-administrator


RichiCoder1:
Superseded by #347.

PR #364, opened 2016-12-20
Improve chocolateyInstall

pascalberger:
Delete MSI after installation and add software name

PR #365, opened 2016-12-20
Remove deprecated functions

pascalberger:
Remove deprecated functions from chocoUninstall and replace aliases with full commands


pascalberger:
@RichiCoder1 Depending on wheter you plan to accept #364 or not, which will remove the MSI after installation :)

RichiCoder1:
There might a future where we have x86 and x64 version of GUI, and when that happens we'll probably switch to remote packages. So I'm onboard with this.

pascalberger:
@RichiCoder1 But thinking of this and that ChocolateyGUI requires a recent Chocolatey version which has auto uninstall feature, I don't think we need the chocolateyUninstall.ps1 at all

Issue #170, opened 2014-09-24
Globalisation/Localisation


gep13:
Do we want to handle this?

If so, turn back on CA1303

RichiCoder1:
Which should at least put everything in place. I think we're actually pretty good about using the resx.

gep13:
@mwallner this is the issue that I think you might want to look at.  Thanks!

mwallner:
Jup, thx - working on it :-) - I'm splitting the dev up into two issues - first one moves English strings to resx - second one adds my German translation. (can assign me if you want) - just a busy week and new StyleCop !#@arrrgh

gep13:
@mwallner are there any rules in particular that you are finding jarring? We can discuss turning other rules off.

gep13:
Thanks to @mwallner this has been taken care of 😄

PR #374, opened 2017-01-15
(GH-170) Globalisation/Localisation

mwallner:
moving all neccessary strings to resx.
(translation from English to other languages in follow-up issues)


mwallner:
the automatic language selection / some kind of dropdown for language selection (maybe in settings?) is still missing - open for any Ideas how / where to implement.

gep13:
Should we default to the current user Locale, with an option to change?

mwallner:
I'd say that's the preferred way, yes

mwallner:
German localization on the way: https://github.com/mwallner/ChocolateyGUI/tree/feature/localization_de - where should we set the 'CultureInfo' ?

RichiCoder1:
CultureInfo should get pulled automagically from the user's machine. It shouldn't, in theory, need to be set. There are ways to override an applications culture on startup if we want to add a manual override though.

RichiCoder1:
For more deets: https://msdn.microsoft.com/en-us/library/system.globalization.cultureinfo.currentuiculture%28v=vs.110%29.aspx

RichiCoder1:
I'll make sure I add strings for all future work, as well as go through and make sure all error messages have Resx sources.

Issue #371, opened 2017-01-08
Add check for useBackgroundService 


gep13:
ChocolateyGUI needs to look in the config file and see if features/useBackgroundService is there and if it is set to true.

Then allow running install/upgrade without trying to seek admin rights.

ferventcoder:
Thanks!

ferventcoder:
cc @RichiCoder1 (I believe we chatted about this awhile back so if this is already done... ;))

RichiCoder1:
@ferventcoder Saaaadly no. Everything is in place to make this check, but I haven't done the work itself. Sadly my actual job went from 1 to 60 just recently.

PR #379, opened 2017-01-20
(GH-371) Respect useBackgroundService Feature

RichiCoder1:


ferventcoder:
**NOTE**: This should only affect the install/upgrade button - all of the others should still require UAC if they would normally.

ferventcoder:
So basically a conditional UAC for that button and everything else a normal UAC.

RichiCoder1:
Hmmm. That's going to be much harder to do then. I'll have to think about that.

Issue #375, opened 2017-01-15
German localization


mwallner:
Strings.de.resx containing German resource strings.
(extends GH-170)

gep13:
@mwallner am I right in saying that this can be closed?

mwallner:
Yup

PR #381, opened 2017-01-27
(GH-375) German localization

mwallner:
first draft of german localization


mwallner:
@gep13 @RichiCoder1 I'm kinda desperate: I can't get the feature to work - the localization is here, I've tried several ways of 'forcing' the app to use the German resx-file .. no chance - any ideas/help on this?

PR #382, opened 2017-01-31
Add check for background service

### LINKED ISSUE #371: Add check for useBackgroundService

RichiCoder1:
Also fix a few typos and back resources.

Closes #371


RichiCoder1:
Had a change to poke at this @ferventcoder?

ferventcoder:
I did, but there is an issue I am going to dive deeper into this week to see whether it stems from this or from the API.

RichiCoder1:
*bump*

ferventcoder:
Gary and I looked deeper into this - what is happening is the licensed.dll seems to run into an error that is super hard to determine - may need to find a way to attach a debugger. What I think it is - because the licensed edition takes a dependency on choco.exe, it's having issues when it is the DLL instead.

RichiCoder1:
Odd. Let me know how I can assist. I believe the chocolatey.lib includes symbols right? Even if not, you should be able to switch Gui's reference to a locally build version, turn off exact source match in debug settings, and then just attached a debugger with `chocolatey` to Gui. The ilmerge might futz with things though, and I'm not sure if and how to make Gui work directly with the source dlls....

RichiCoder1:
*bump*

ferventcoder:
There have been some fixes into chocolatey.lib and the extension that may fix this. Getting ready to try this again.

RichiCoder1:
Excellent! I'd merge mainline develop into this as it just updated all the the packages (including lib)

ferventcoder:
Can I push to this PR?

RichiCoder1:
Absolutely

ferventcoder:
Should I wait until we get the other through? I prefer to reduce merge conflicts and so much has changed.

RichiCoder1:
#438 is merged and cleaned up. You should be good to go now.

RichiCoder1:
Brought this branch back up to Develop. Builds and runs just fine, though I'm not sure how much of the BP functionality is still preserved.

ferventcoder:
BP?

RichiCoder1:
Background service stuff. BS doesn't have the same ring haha.

ferventcoder:
Ah true!

ferventcoder:
Superseded by #451

PR #383, opened 2017-02-01
Fix xaml culture

RichiCoder1:
/cc @mwallner 

Certain things like date times will still have to be manually tweaked, but this should cover the issue in 90% of cases. This *should* fix resxs. Still need to test more.

Related; #381


mwallner:
Well, it looks promising .. just figured out I probably should have used Resources.resx from the beginning instead of creating a 'new' Strings.resx :-( .. when I copy the content of Strings.resx to Resources.resx and move Strings.de.resx to Resources.de.resx - localization is working now! https://twitter.com/schusterfredl/status/826689323647004672

Issue #384, opened 2017-02-01
move localization/resource strings from Strings.resx to Resources.resx


mwallner:
seemingly the 'automatic' localization only works for Resources.resx ?

gep13:
@mwallner is this work still to be done?

mwallner:
Nope, already in GH-386

PR #385, opened 2017-02-01
(GH-384) move localization/resource strings from Strings.resx to Resources.resx

mwallner:
seemingly the 'automatic' localization only works for Resources.resx
* requires GH-383


mwallner:
moved to #386

Issue #405, opened 2017-02-14
Text/Header strings not prepared to be translated


AdmiringWorm:
While working on a translation to Norwegian Bokmål I noticed that there was certain parts that haven't yet been prepared to be translated.

*Current findings*
The current parts that I found that haven't been prepared yet.
- Local/Remote package context menu
- Disabled Header in the SettingsView_Source

Places referencing the wrong string
- PackageView pin button is referencing the install button text.

I've fixed this in a local commit, so if interested I can open a PR for it.

mwallner:
I'd like to see a PR :+1:

AdmiringWorm:
@mwallner PR submitted: #406

RichiCoder1:
Closed by #406

gep13:
@AdmiringWorm thank you for submitting this, appreciate you taking the time to look at it!

AdmiringWorm:
@gep13 no problem, I probably missed some strings somewhere, as well as strings related to exceptions and logging was not added in #406. Mainly because, logging shouldn't be translated IMO, and didn't know which exceptions was displayed to the user so basically ignored those.

PR #406, opened 2017-02-14
(GH-405) Moved hard coded text strings to resource file

AdmiringWorm:
A few text and header attributes in the xaml files was not moved to the Resource.resx file.
this PR fixes that.
It also changes ResXFileCodeGenerator to PublicResXFileCodeGenerator so that if new strings are added to the Resource.rex file or the designer is regenerated it will make the properties public (which is needed to be able to use them in xaml).

#405

### DIFF
--- a/Source/ChocolateyGui/Resources/Controls.xaml
+++ b/Source/ChocolateyGui/Resources/Controls.xaml
@@ -661,7 +661,7 @@
                 </MultiBinding>
             </MenuItem.Visibility>
         </MenuItem>
-        <MenuItem Header="Unpin" Icon="{StaticResource UacIcon}"
+        <MenuItem Header="{x:Static lang:Resources.Controls_PackagesContextMenuUnpin}" Icon="{StaticResource UacIcon}"
                   Command="{commands:DataContextCommandAdapter Unpin}">
             <MenuItem.Visibility>
                 <MultiBinding Converter="{StaticResource BooleanToVisibility}">
--- a/Source/ChocolateyGui/Resources/Controls.xaml
+++ b/Source/ChocolateyGui/Resources/Controls.xaml
@@ -649,10 +649,10 @@
     <ContextMenu x:Key="PackagesContextMenu"
                  DataContext="{Binding PlacementTarget.DataContext, RelativeSource={RelativeSource Self}}"
                  d:DataContext="{d:DesignInstance Type=items:PackageViewModel}">
-        <MenuItem Header="Install" Icon="{StaticResource UacIcon}"
+        <MenuItem Header="{x:Static lang:Resources.Controls_PackagesContextMenuInstall}" Icon="{StaticResource UacIcon}"
                   Visibility="{Binding IsInstalled, Converter={StaticResource BooleanToVisibility}, ConverterParameter=True}"
                   Command="{commands:DataContextCommandAdapter Install}" />
-        <MenuItem Header="Pin" Icon="{StaticResource UacIcon}"
+        <MenuItem Header="{x:Static lang:Resources.Controls_PackagesContextMenuPin}" Icon="{StaticResource UacIcon}"
                   Command="{commands:DataContextCommandAdapter Pin}">
             <MenuItem.Visibility>
                 <MultiBinding Converter="{StaticResource BooleanToVisibility}">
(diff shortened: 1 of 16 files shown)


AdmiringWorm:
ops, what a mistake to make.
I'll probably end up squashing everything before it's merge, but still I'll rebase the changes for know and correct my mistake.

AdmiringWorm:
I've fixed the mistake, as well as added quite a few new resource strings from the C# Code.

Source/ChocolateyGui/Resources/Controls.xaml
```diff
@@ -670,16 +670,16 @@
...
-        <MenuItem Header="Uninstall" Icon="{StaticResource UacIcon}"
+        <MenuItem Header="{x:Static lang:Resources.Controls_PackagesContextMenuUninstall}" Icon="{StaticResource UacIcon}"
                   Visibility="{Binding IsInstalled, Converter={StaticResource BooleanToVisibility}}"
                   Command="{commands:DataContextCommandAdapter Uninstall}" />
-        <MenuItem Header="Reinstall" Icon="{StaticResource UacIcon}"
+        <MenuItem Header="{x:Static lang:Resources.Controls_PackagesContextMenuReinstall}" Icon="{StaticResource UacIcon}"
                   Visibility="{Binding IsInstalled, Converter={StaticResource BooleanToVisibility}}"
                   Command="{commands:DataContextCommandAdapter Reinstall}" />
-        <MenuItem Header="Update" Icon="{StaticResource UacIcon}"
+        <MenuItem Header="{x:Static lang:Resources.Controls_PackagesContextMenuUpdate}" Icon="{StaticResource UacIcon}"
```
RichiCoder1: Hmm. Maybe at some point we should come up with a strategy so we don't duplicate resource strings needlessly?

Source/ChocolateyGui/Services/ChocolateyRemotePackageService.cs
```diff
@@ -242,7 +282,9 @@ public ValueTask<bool> RequiresElevation()
         public void Dispose()
         {
             _logStream?.Dispose();
-            _wampChannel?.Close("Exiting", new GoodbyeDetails { Message = "Exiting" });
+            _wampChannel?.Close(
+                Resources.ChocolateyRemotePackageService_Exiting,
```
RichiCoder1: Actually, these need to stay hardcoded magic strings. I'll move them to a const later.

Source/ChocolateyGui/Services/ChocolateyRemotePackageService.cs
```diff
@@ -283,7 +325,9 @@ private Task Initialize(bool requireAdmin = false)
                     _isInitialized = false;
                     _logStream.Dispose();
                     _logStream = null;
-                    _wampChannel.Close("Escalating", new GoodbyeDetails { Message = "Escalating" });
+                    _wampChannel.Close(
+                        Resources.ChocolateyRemotePackageService_Escalating,
+                        new GoodbyeDetails { Message = Resources.ChocolateyRemotePackageService_Escalating });
```
RichiCoder1: Same here. These are intentionally magical.

Source/ChocolateyGui/ViewModels/RemoteSourceViewModel.cs
```diff
@@ -231,9 +232,9 @@ protected override void OnInitialize()
                 MessageBox.Show(
```
RichiCoder1: Oh wow, this is old. I'll need to rip this out in a seperate PR.

Source/ChocolateyGui/ViewModels/SourcesViewModel.cs
```diff
@@ -90,7 +90,7 @@ protected override void OnViewReady(object view)
         {
             Observable.FromEventPattern<PropertyChangedEventArgs>(this, nameof(PropertyChanged))
                 .Where(p => p.EventArgs.PropertyName == nameof(ActiveItem))
-                .Subscribe(p => DisplayName = $"Source - {ActiveItem.DisplayName}");
+                .Subscribe(p => DisplayName = $"Source - {ActiveItem?.DisplayName}");
```
RichiCoder1: 👍

Source/ChocolateyGui/Resources/Controls.xaml
AdmiringWorm: Could perhaps seperate the strings out into their own Resource file, while just having common strings in the Resource.resx file?

Source/ChocolateyGui/Services/ChocolateyRemotePackageService.cs
AdmiringWorm: The Message property as well?

Source/ChocolateyGui/ViewModels/SourcesViewModel.cs
AdmiringWorm: I got so annoyed by getting a null exception every time I closed the program, so just had to add it.

Source/ChocolateyGui/Services/ChocolateyRemotePackageService.cs
RichiCoder1: Yup. Ditto for other.

Source/ChocolateyGui/Resources/Controls.xaml
RichiCoder1: Hmm. I think we ran into an issue where localization wasn't happening for other Resxs? Maybe have been a misconfiguration though. I'd be for a Resources_Common.resx or something like that.

Source/ChocolateyGui/ViewModels/SourcesViewModel.cs
RichiCoder1: Yup. I have this fixed properly on another outstanding branch.

AdmiringWorm: Should probably revert this change then, or?

RichiCoder1: Go ahead and keep it. I'll fix it up in my other PR, as that's looking like it won't get merged for a few days yet still.

Source/ChocolateyGui/Resources/Controls.xaml
AdmiringWorm: I can probably look into that, but that probably deserves its own PR though.

AdmiringWorm: I don't really see how it would add much complexity.
IMO it could rather be easier to maintain if strings are located in their own resource file. (Common.resx, PackageView.resx, etc.)

However, this may come down to preference though.

RichiCoder1: I'd be up for having a handful of resource category files (common, errors, UI, images), as I've done that previously and it's worked well. Per-component would be excessive IMO however.

RichiCoder1:
Awesome work :)

PR #410, opened 2017-02-20
Localization/missing translations

magol:
Make it possible to localizate the description strings in settings and On/Off
Remove some unused resource strings

### DIFF
--- a/Source/ChocolateyGui/Properties/Resources.Designer.cs
+++ b/Source/ChocolateyGui/Properties/Resources.Designer.cs
@@ -1064,6 +1064,24 @@ namespace ChocolateyGui.Properties {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Off.
+        /// </summary>
+        public static string SettingsView_Off {
+            get {
+                return ResourceManager.GetString("SettingsView_Off", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to On.
+        /// </summary>
+        public static string SettingsView_On {
+            get {
+                return ResourceManager.GetString("SettingsView_On", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to Options.
         /// </summary>
         public static string SettingsView_Options {
--- a/Source/ChocolateyGui/Properties/Resources.Designer.cs
+++ b/Source/ChocolateyGui/Properties/Resources.Designer.cs
@@ -61,15 +61,6 @@ namespace ChocolateyGui.Properties {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Argument can&apos;t be null or empty..
-        /// </summary>
-        public static string Argument_cant_be_null_or_empty {
-            get {
-                return ResourceManager.GetString("Argument_cant_be_null_or_empty", resourceCulture);
-            }
-        }
-        
-        /// <summary>
         ///   Looks up a localized string similar to Unhandled Exception.
         /// </summary>
         public static string Bootstrapper_UnhandledException {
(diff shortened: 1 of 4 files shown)


mwallner:
@gep13 @RichiCoder1 how should we proceed when strings are deleted from Resources.resx - file?
(other languages should be updated somehow..)

magol:
@mwallner Yes, I was thinking about same thing. There must be a method how to do this.

The method for removing a string should be pretty simple, the string must be removed from all languages.

The methodology is to add strings are a little bit more uncertain. One suggestion is to add the strings for all languages, but let them be in English. Maybe add a comment so that translators can easily find them.

magol:
@mwallner I have now removed the unused resource strings from DE to.

Source/ChocolateyGui/ViewModels/SettingsViewModel.cs
```diff
@@ -270,6 +272,26 @@ public void Back()
...
             Sources.AddRange(sources);
+        }
+
+        private void LocalizateFeaturesDescription(ChocolateyFeature feture)
+        {
+            var key = string.Format("SettingsViewModel_FeatureDescription_{0}", feture.Name);
+            var description = Resources.ResourceManager.GetString(key, Resources.Culture);
+            if (!string.IsNullOrWhiteSpace(description))
+            {
+                feture.Description = description;
```
mwallner: @gep13 talking CI - could we implement something that would cause an Warning/Error on build? - this way we could at least ensure all settings strings were available/translated on build date?

Source/ChocolateyGui/ViewModels/SettingsViewModel.cs
```diff
@@ -270,6 +272,26 @@ public void Back()
 
             var sources = await _packageService.GetSources();
             Sources.AddRange(sources);
+        }
+
+        private void LocalizateFeaturesDescription(ChocolateyFeature feture)
```
gep13: Can we change `feture` to `feature`?

Source/ChocolateyGui/ViewModels/SettingsViewModel.cs
gep13: @mwallner yes, this is definitely something that can be done.  Personally, I am happy to do this, but last time I checked, there were some un related warnings in the build step, which would ultimately cause the build to fail.  If everyone is happy to proceed with warnings as errors, then this can certainly be enabled.

mwallner: treating warnings as errors (at least in 'Release' configuration) sounds like a good idea!
- shouldn't be to complicated to clean up the current warnings?

RichiCoder1:
Sounds like there's still some feedback? Once @gep13 approves I'll do a sanity check and merge.

Source/ChocolateyGui/ViewModels/SettingsViewModel.cs
magol: @gep13 Fixed :-)

RichiCoder1:
Hmm. I can't help but think that feature/setting description localization is something that should be pushed upstream into Choco. /cc @ferventcoder ?

ferventcoder:
It does make sense to do some localization in Choco itself. I'm wondering how best to make that available.

magol:
@richicoder1 ok,  i remove the localizarion of the description

magol:
@richicoder1 I have removed the localization of descriptions now.

What should I do with the removed code? Should I create a PR with that code, if we in the future select to have the localization in ChocolateyGUI instead of coco?

RichiCoder1:
Maybe keep it in it's own branch, but I wouldn't make a PR until we come to a decision.

magol:
@RichiCoder1  ok.
I think this is ready for merge now

PR #400, opened 2017-02-10
Translate to Swedish

magol:
#396

Issue #396, opened 2017-02-05
Globalisation - Swedish


gep13:
@devlead @patriksvenson following the work that @mwallner has done to get support for the German langauge, would you be interested in providing support for Swedish?  Thanks!

magol:
Jag hjälper gärna till att översätta till svenska 😊

I'm want to help to translate to Swedish

gep13:
>@magol said...
I'm want to help to translate to Swedish

That is great news!

Perhaps @mwallner could provide a brief overview of how to approach adding another locale into the project?

mwallner:
nice! - I'm going to write a little 'HOWTO- do translation for ChocolateyGUI' - but It'll take 1-2 days...

PR #412, opened 2017-02-21
(GH-396) Added Swedish translation

magol:
I made a mistake in #400, so I restart with this new PR for the Swedish translation.

I have taken into account the review comments from the old PR.  

#396

### DIFF
--- a/Source/ChocolateyGui/ChocolateyGui.csproj
+++ b/Source/ChocolateyGui/ChocolateyGui.csproj
@@ -288,6 +288,11 @@
       <AutoGen>True</AutoGen>
       <DesignTime>True</DesignTime>
     </Compile>
+    <Compile Include="Properties\Resources.sv.Designer.cs">
+      <AutoGen>True</AutoGen>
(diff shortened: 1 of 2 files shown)


Source/ChocolateyGui/ChocolateyGui.csproj
```diff
@@ -495,6 +495,7 @@
       <Generator>PublicResXFileCodeGenerator</Generator>
       <LastGenOutput>Resources.de.Designer.cs</LastGenOutput>
     </EmbeddedResource>
+    <EmbeddedResource Include="Properties\Resources.sv.resx" />
```
RichiCoder1: I think you're right @mwallner. Why doesn't this look like the other ones? @magol Could you update this entry to look like the other two resource file entries?

magol: @RichiCoder1 Oups, I have fixed that now :-)

RichiCoder1:
Awesome work!

PR #438, opened 2017-05-14
Refactor Subprocess Structure

RichiCoder1:
This PR accomplishes two major things: 

* It removes the last vestiges of the old style way of how we were doing a number of things in Gui behind the scenes. Also cleans up our async primitives to use the seriously high quality VS.Threading package.
* Replaces the buggy and ill convinced WAMP IPC protocol with a much more solid WCF IPC layer. Mostly for my santity, but it has had the nice side effect of crashing less and being ever so slightly faster.

Other things of note:
* I redid how we're copying Subprocess into Gui's output. It's much more sane now, (and gets rid of an app.config issue that was there previously). It also means that Main no longer references Subprocess, but instead a shared project to get Dtos.
* ChocolateyGui (main) no longer has a dependency on `chocolatey.lib` (ha). All the choco specific bits have been abstracted and contained in Subprocess.

Lastly, one thing that @ferventcoder and I will both appreciate, you can now start ChocolateyGui.Subprocess before, and independent of, ChocolateyGui. This makes it muuuuch easier to debug Subprocess as you can just set both projects as startup projects in VS (with Subprocess starting first), allowing you to simultaneously attach to both.


RichiCoder1:
AppVeyor is agree w/ me because of https://github.com/cake-contrib/Cake.Recipe/issues/97

ferventcoder:
I'm all about fixing stuff. Let me take a looksy

Source/ChocolateyGui.Shared/Models/Package.cs
```diff
@@ -5,71 +5,134 @@
...
+        [DataMember]
+
+        public bool IsAbsoluteLatestVersion { get; set; }
+
+        [DataMember]
+
         public bool IsInstalled { get; set; }
 
+        [DataMember]
+
```
ferventcoder: looks like extra lines inserted here. Issues with flipping this away or to UTF8 usually causes this.

Source/ChocolateyGui.Shared/Models/PackageOperationResult.cs
```diff
@@ -5,21 +5,25 @@
...
         {
             Successful = true
         };
 
+        [DataMember]
         public bool Successful { get; set; }
 
-        public IReadOnlyList<string> Messages { get; set; } = new string[0];
+        [DataMember]
+        public string[] Messages { get; set; } = new string[0];
```
ferventcoder: So it has to be an array? Can't be a `List<string>`? (I'm pretty sure IList won't work here).

Source/ChocolateyGui.Shared/Models/SemanticVersion.cs
```diff
@@ -0,0 +1,337 @@
...
+            {
+#pragma warning disable SA1407 // Arithmetic expressions must declare precedence
+                hashCode = hashCode * 4567 + SpecialVersion.GetHashCode();
+#pragma warning restore SA1407 // Arithmetic expressions must declare precedence
+            }
+
+            return hashCode;
+        }
+    }
+}
```
ferventcoder: Why did you need to implement this? I'd prefer using it directly out of chocolatey.dll.

ferventcoder: Unless it is being internalized, which I think we fixed a long time ago.

Source/ChocolateyGui.Subprocess/ChocolateyGui.Subprocess.csproj
```diff
@@ -38,54 +37,21 @@
...
-      <HintPath>..\packages\Newtonsoft.Json.10.0.2\lib\net45\Newtonsoft.Json.dll</HintPath>
-    </Reference>
-    <Reference Include="Newtonsoft.Msgpack, Version=1.0.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\Newtonsoft.Msgpack.0.1.10\lib\net45\Newtonsoft.Msgpack.dll</HintPath>
-      <Private>True</Private>
-    </Reference>
-    <Reference Include="Nito.AsyncEx, Version=4.0.1.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\Nito.AsyncEx.4.0.1\lib\net45\Nito.AsyncEx.dll</HintPath>
+    <Reference Include="Microsoft.VisualStudio.Threading, Version=15.3.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.VisualStudio.Threading.15.3.20\lib\net45\Microsoft.VisualStudio.Threading.dll</HintPath>
```
ferventcoder: Weird name...

Source/ChocolateyGui.Subprocess/packages.config
```diff
@@ -1,34 +1,17 @@
...
   <package id="System.Reactive.Core" version="3.1.1" targetFramework="net452" />
   <package id="System.Reactive.Interfaces" version="3.1.1" targetFramework="net452" />
   <package id="System.Reactive.Linq" version="3.1.1" targetFramework="net452" />
-  <package id="System.Reactive.PlatformServices" version="3.1.1" targetFramework="net452" />
-  <package id="System.Reactive.Windows.Threading" version="3.1.1" targetFramework="net452" />
-  <package id="System.Threading.Tasks.Dataflow" version="4.7.0" targetFramework="net452" />
-  <package id="System.ValueTuple" version="4.3.1" targetFramework="net452" />
-  <package id="WampSharp" version="1.2.5.27-beta" targetFramework="net452" />
-  <package id="WampSharp.Fleck" version="1.2.5.27-beta" targetFramework="net452" />
-  <package id="WampSharp.NewtonsoftJson" version="1.2.5.27-beta" targetFramework="net452" />
```
ferventcoder: That really reduced some dependencies!!

Source/ChocolateyGui/Services/ChocolateyRemotePackageService.cs
```diff
@@ -337,124 +333,49 @@ private Task Initialize(bool requireAdmin = false)
...
-                                case StreamingLogLevel.Fatal:
-                                    powerShellLineType = PowerShellLineType.Error;
-                                    break;
-                                default:
-                                    powerShellLineType = PowerShellLineType.Output;
-                                    break;
-                            }
-
-                            _progressService.WriteMessage(message.Message, powerShellLineType);
-                        });
```
ferventcoder: You don't need any of this now? I'm guessing you have a logger set up.

Source/ChocolateyGui/Services/ChocolateyRemotePackageService.cs
```diff
@@ -471,12 +392,37 @@ private void LogError()
...
+                var channel =
+                    new DuplexChannelFactory<IIpcChocolateyService>(context, binding, endpoint).CreateChannel();
+                channel.Register();
+                return channel;
+            }
+            catch (Exception ex)
+            {
+                Logger.Fatal(ex, "Failed to create client channel to server.");
+                throw;
+            }
```
ferventcoder: 👍

Source/ChocolateyGui/Services/ServiceCallbackHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+                case StreamingLogLevel.Fatal:
+                    powerShellLineType = PowerShellLineType.Error;
+                    break;
+                default:
+                    powerShellLineType = PowerShellLineType.Output;
+                    break;
+            }
+
+            _progressService.WriteMessage(message.Message, powerShellLineType);
+        }
```
ferventcoder: And there it is.

Source/ChocolateyGui/ViewModels/SettingsViewModel.cs
```diff
@@ -251,7 +234,10 @@ public void Back()
             WireUpConfig();
 
             var features = await _packageService.GetFeatures();
-            Features.AddRange(features);
+            foreach (var feature in features)
+            {
+                Features.Add(feature);
+            }
```
ferventcoder: Optimization?

Source/ChocolateyGui.Shared/Models/PackageOperationResult.cs
RichiCoder1: I honestly don't know haha. It was yelling at me for other things. Will try again.

Source/ChocolateyGui.Shared/Models/SemanticVersion.cs
RichiCoder1: Nope, this was because I pulled chocolatey.lib out of chocolateygui (so it's only the in the subprocess). However, there's 0 reason I can't just put it back, especially since I'm just using a model.

Source/ChocolateyGui.Subprocess/ChocolateyGui.Subprocess.csproj
RichiCoder1: Suprisingly, it's well names. It the Microsoft Visual Studio Team's threading library. It's actually really awesome (hence why I swapped Nito.AsyncEx with it.)

Source/ChocolateyGui.Subprocess/packages.config
RichiCoder1: That was 90% of the reason for this change. WampSharp just had some many superflous dependencies and issues.

Source/ChocolateyGui/ViewModels/SettingsViewModel.cs
RichiCoder1: Actually could be optimized further. I don't what library I pulled in had it, but ObservableCollection (which backs Features, Settings, and Sources on this page) doesn't have an `AddRange`, so I had to fall back to good old fashion add.

RichiCoder1:
I'll address the interface issues in a later PR. Should be good to go.

PR #441, opened 2017-06-06
(GH-436) Remove hacks for accessing chocolatey internals.

### LINKED ISSUE #436: Remove the Hacks where you access Chocolatey's SimpleInjector.Container
Is there a chance we can remove this? 
https://github.com/chocolatey/ChocolateyGUI/blob/13d0d0f08d5b0eabc108a21f42337b8a7f05a76e/Source/ChocolateyGui.Subprocess/Hacks.cs#L47-L87

Took me awhile to discover this one - accessing internals from something else can lead to really hard to find bugs. Here's a hint on the bug - https://github.com/chocolatey/choco/commit/61cd0847d4198828965626f32c4a3461bb4ee604

> This fixes calls to `RegisterContainerComponents` complaining that components cannot be registered after calls to `GetInstance`, etc.

### Details
1. Once the Container is loaded, you cannot add new things to it.
1. [`ChocolateyService.GetSources()` is one of the first things to run]( https://github.com/chocolatey/ChocolateyGUI/blob/13d0d0f08d5b0eabc108a21f42337b8a7f05a76e/Source/ChocolateyGui.Subprocess/ChocolateyService.cs#L377) in the subprocess - note it makes a call to the Hacks file which accesses `SimpleInjector.Container`.
1. Later, [`Lets.GetChocolatey()` is called](https://github.com/chocolatey/ChocolateyGUI/blob/13d0d0f08d5b0eabc108a21f42337b8a7f05a76e/Source/ChocolateyGui.Subprocess/ChocolateyService.cs#L51) and then tries to load licensed code registration into the container, however it can't because the container has already been accessed. 
1. And then nothing works like you would expect it to.

### Workaround
The fix I have for this now is to call `Lets.GetChocolatey()` in `Program.Main()` when the process is setting up so that everything gets registered properly prior to calling `GetInstance`.

### Fix
I may not fully be aware of the reasoning here, is there a reason this is being used over setting those items up in a container specifically for the GUI?

I spent more time that I want to admit wrestling with why I could not get the Licensed code to set up properly.

### Other Thoughts
I did spend quite a bit of time making the lib code a bit more robust, so at least there is that. 

I'll be pushing back some PRs once I get all of this cleaned up.

RichiCoder1:
Fixes #436.

### DIFF
--- a/Source/ChocolateyGui.Shared/ChocolateyGui.Shared.csproj
+++ b/Source/ChocolateyGui.Shared/ChocolateyGui.Shared.csproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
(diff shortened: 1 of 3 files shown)


RichiCoder1:
Gotta love deleting code :)

Issue #445, opened 2017-06-08
ChocolateyGUI doesn't display VersionInfo from internal feeds


mwallner:
(LocalSourceView)
The column 'latest version' is empty for every package that doesn't exist on the public feed - even if is on an internal (ProGet) feed.
- if the public feed is disabled, no package has a 'latest version', hence no updates may be done via the UI

version is current develop (d1d57022c0dc6e0fa822edebae4a0715fe7b8a9e)

mwallner:
fixed with GH-446

PR #446, opened 2017-06-08
(GH-445) Fix outdated check

RichiCoder1:
Switch to implentation of outdated that mirrors chocolatey.


Source/ChocolateyGui.Subprocess/ChocolateyService.cs
```diff
@@ -70,6 +72,7 @@ public Task<bool> IsElevated()
             var operationContext = OperationContext.Current;
             using (await Lock.ReadLockAsync())
             {
+                var originalConfig = Config.get_configuration_settings().deep_copy();
```
mwallner: I'd like to get @ferventcoder 's opinion on this

ferventcoder: Looks legit to me.

PR #451, opened 2017-06-12
(GH-371) Enable self-service mode

### LINKED ISSUE #371: Add check for useBackgroundService

ferventcoder:
When using self-service installs, a feature in Chocolatey for Business,
it provides a background agent that actually performs the
installations. Chocolatey itself handles passing through the proper
information to the background service so that installations can occur.
The GUI should notice the feature is switched on and adjust to allow
non-administrators to run certain commands in the GUI without
attempting to elevate.

Supersedes #382
Closes #371

### DIFF
--- a/Source/ChocolateyGui/ChocolateyGui.csproj
+++ b/Source/ChocolateyGui/ChocolateyGui.csproj
@@ -233,7 +233,7 @@
     <Compile Include="Providers\ChocolateyConfigurationProvider.cs" />
     <Compile Include="Providers\ChocolateyCustomSchemeProvider.cs" />
     <Compile Include="Providers\ChocoRequestHandler.cs" />
-    <Compile Include="Providers\ElevationStatusProvider.cs" />
+    <Compile Include="Elevation.cs" />
     <Compile Include="Providers\IChocolateyConfigurationProvider.cs" />
     <Compile Include="Providers\IPlatformProvider.cs" />
     <Compile Include="Providers\IVersionNumberProvider.cs" />
--- a/Source/ChocolateyGui/ChocolateyGui.csproj
+++ b/Source/ChocolateyGui/ChocolateyGui.csproj
@@ -183,6 +183,7 @@
       <Generator>MSBuild:Compile</Generator>
       <SubType>Designer</SubType>
     </Page>
+    <Compile Include="Base\ConnectionClosedException.cs" />
     <Compile Include="Controls\InternetImage.xaml.cs">
       <DependentUpon>InternetImage.xaml</DependentUpon>
     </Compile>
--- a/Source/ChocolateyGui/ChocolateyGui.csproj
+++ b/Source/ChocolateyGui/ChocolateyGui.csproj
@@ -60,8 +60,9 @@
     <Reference Include="Caliburn.Micro.Platform.Core, Version=3.0.3.0, Culture=neutral, PublicKeyToken=8e5891231f2ed21f, processorArchitecture=MSIL">
       <HintPath>..\packages\Caliburn.Micro.3.1.0\lib\net45\Caliburn.Micro.Platform.Core.dll</HintPath>
     </Reference>
-    <Reference Include="chocolatey, Version=0.10.6.1, Culture=neutral, PublicKeyToken=79d02ea9cad655eb, processorArchitecture=MSIL">
-      <HintPath>..\packages\chocolatey.lib.0.10.6.1\lib\chocolatey.dll</HintPath>
+    <Reference Include="chocolatey, Version=0.10.7.0, Culture=neutral, PublicKeyToken=79d02ea9cad655eb, processorArchitecture=MSIL">
+      <HintPath>..\packages\chocolatey.lib.0.10.7\lib\chocolatey.dll</HintPath>
+      <Private>True</Private>
     </Reference>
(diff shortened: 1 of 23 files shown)


ferventcoder:
There is still something not quite right here: 

image


Then the connection between the GUI and the subprocess is super erratic. Is this a known issue?

ferventcoder:
I think I found an issue with the background service that didn't bubble up to the GUI at all. 

"[Fatal] Calls to Chocolatey only are allowed. Attempted call to 'C:\vagrant\resources\ChocolateyGiu\Subprocess\ChocolateyGui.Subprocess.exe'. This has been logged for an administrator to review."

ferventcoder:
But that means progress!

ferventcoder:
I'm guessing the fatal error kills the GUI Subprocess.

ferventcoder:
Still getting issues with connections and this as well:

image

ferventcoder:
Looking at the logs it appears things are working appropriately, but something causes it to kill after about 2-3 seconds. Running it with straight choco and it completes successfully.

I wonder if it is because GUI uses `net.tcp` and background mode uses `net.pipe` and something there is incompatible?

ferventcoder:
No, it looks like the agent returns a message that makes the GUI think it is complete.

ferventcoder:
@RichiCoder1 @gep13 appveyor issues don't seem related - it's a cake error.

RichiCoder1:
Looks like it's just a StyleCop error: `ViewModels\LocalSourceViewModel.cs(22,1): error SA1210: Using directives must be ordered alphabetically by the namespaces` 

That run issue is a Xaml issue (that I could have sworn we already fixed.)

RichiCoder1:
`[Fatal]` generally means an exception bubbled all the way up the host and was unhandled. So you can assume that anything with that tag was, well, fatal.

Source/ChocolateyGui/ChocolateyGui.csproj
```diff
@@ -233,7 +234,7 @@
     <Compile Include="Providers\ChocolateyConfigurationProvider.cs" />
     <Compile Include="Providers\ChocolateyCustomSchemeProvider.cs" />
     <Compile Include="Providers\ChocoRequestHandler.cs" />
-    <Compile Include="Providers\ElevationStatusProvider.cs" />
+    <Compile Include="Elevation.cs" />
```
mwallner: this will probably lead to a StyleCop err?

mwallner:
@RichiCoder1 I think we can merge this into develop - will create another intermediate build and do some testing - is that ok with you?

ferventcoder:
As long as all of the xaml is good to go, I'm happy with that - we'll change this from closes though.

RichiCoder1:
I suppose most all the remaining work is in Choco.Lib and the integration layer and those can be fixed in future PRs.

ferventcoder:
There is still some work here to be completed - like the issue with something shutting down the self-service before it completes the install 

And ensuring those values do get set in source options.

RichiCoder1:
@ferventcoder Could you elaborate? I'm not sure I follow.

RichiCoder1:
Good heavens above, rebasing can go die in a fire.

ferventcoder:
Let me just say the rebase was especially painful the last time I did it

ferventcoder:
Source options like skip proxy and allow self-service are not actually saved, so I may have missed a setting somewhere. It looks like they are set, but then they are not.

ferventcoder:
And the self-service aspect still does not actually work for a non-admin - it dies when attempting to do the install. Not sure what causes it just yet.

ferventcoder:
A couple of additional clean up items on this side, but this will work perfectly once the following are released:

* chocolatey-agent v 0.7.0
* chocolatey.extension 1.11.0

I will add those additional items here today.

mwallner:
bump @RichiCoder1 :-)

RichiCoder1:
I believe @ferventcoder was just wrapping up the fixes for this.

ferventcoder:
Was there something left for me here?

ferventcoder:
I know that there was an issue with saving the self-service and bypass proxy settings, but I wasn't sure what I missed.

RichiCoder1:
I suppose the todo items (deploy fixed version of agent and extension) are technically independent of this PR. Have we confirmed that it works w/ agent now?

RichiCoder1:
Once we merge, I'll do follow up PRs to make sure everything is polished.

ferventcoder:
I've confirmed it!

ferventcoder:
I blame you ;)

ferventcoder:
This is so much win right here.

mwallner:
well see x_x - I'll hand out this version to a couple of beta-testers tomorrow ;-)

mwallner:
AppVeyor still blows up because of Cake/StyleCop .. will fix tomorrow - TeamCity is fine.

ferventcoder:
Did you see the only apply background service to non-admins coming in Chocolatey Licensed v1.12?

RichiCoder1:
Ahh, this explains that explosion I saw.

Issue #432, opened 2017-04-26
Limit non-admin list to self service only.


lukemgriffith:
Currently with feature useBackgroundServiceWithSelfServiceSourcesOnly enabled, a user without admin permissions can do choco list and return items from all configured sources by default, including packages from sources that are not configured for self service.

Would it be possible for non elevated users to only have self service packages returned, as its the only items they can actually install.

ferventcoder:
If the background service is on and the user is a non-admin that is.

RichiCoder1:
I can look into this after we do #382. Would chocolatey gui do anything? Or would choco.lib be smart enough to only pass back relevant sources & packages?

ferventcoder:
@RichiCoder1 Chocolatey.lib would be smart enough only to pass back relevant sources.

ferventcoder:
I'll start adding the details for this feature

RichiCoder1:
@ferventcoder Is this feature complete?

ferventcoder:
I started on this, there is still one component missing here.

gep13:
@ferventcoder @RichiCoder1 should we bump this to the next release of Chocolatey GUI?

ferventcoder:
Yes, please.

gep13:
@ferventcoder am I right in saying that this will be included in the change that I am away to make for using `source_list` rather then reading the available sources directly from the Chocolatey Configuration file?

ferventcoder:
@gep13 @lukemgriffith I want to say this is covered by the change to not see admin only - that's where we've decided to hide those sources that would be admin only. Non-admins would see self-service and normal sources.

ferventcoder:
I want to say that's where this conversation ended up.

PR #459, opened 2017-08-29
(GH-455)(GH-432) Sources updates / Chocolatey 0.10.8 updates

### LINKED ISSUE #455: Source Configuration - Seeing and Setting Bypass Proxy and Allow Self-Service on a Source doesn't work
This is not pulling the data properly, nor is it setting it. Something is not getting set correctly in the binding to make this setting take. I am somewhat limited on WPF to see what it may be.

ferventcoder:
Related to #432 
Closes #455 

This includes some changes that will only work with Chocolatey v0.10.8.

### DIFF
--- a/Source/ChocolateyGui.Shared/ChocolateyGui.Shared.csproj
+++ b/Source/ChocolateyGui.Shared/ChocolateyGui.Shared.csproj
@@ -33,6 +33,7 @@
   <ItemGroup>
     <Reference Include="chocolatey, Version=0.10.7.0, Culture=neutral, PublicKeyToken=79d02ea9cad655eb, processorArchitecture=MSIL">
       <HintPath>..\packages\chocolatey.lib.0.10.7\lib\chocolatey.dll</HintPath>
+      <SpecificVersion>False</SpecificVersion>
       <Private>True</Private>
     </Reference>
     <Reference Include="log4net, Version=1.2.13.0, Culture=neutral, PublicKeyToken=669e0ddf0bb1aa2a, processorArchitecture=MSIL">
--- a/Source/ChocolateyGui.Shared/ChocolateyGui.Shared.csproj
+++ b/Source/ChocolateyGui.Shared/ChocolateyGui.Shared.csproj
@@ -31,10 +31,8 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="chocolatey, Version=0.10.7.0, Culture=neutral, PublicKeyToken=79d02ea9cad655eb, processorArchitecture=MSIL">
-      <HintPath>..\packages\chocolatey.lib.0.10.7\lib\chocolatey.dll</HintPath>
-      <SpecificVersion>False</SpecificVersion>
-      <Private>True</Private>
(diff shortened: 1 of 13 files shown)


ferventcoder:
@RichiCoder1 so I'm seeing some weird behavior where a boolean is flagged true, but once it comes over from the subprocess to ChocolateyService for update, it resets to false. This is what I think is why the new settings for sources are not getting applied. I haven't dug in too much, but does that ring a bell at all for you?

ferventcoder:
This will fail the build until there is a Chocolatey v0.10.8.

ferventcoder:
Added Chocolatey 0.10.8 and some other fixes to appease the builds. All green!!

ferventcoder:
@gep13 can you take a closer look at this?

ferventcoder:
I'm happy to squash/rebase prior to merge as well.

Source/ChocolateyGui.Shared/ChocolateyGui.Shared.csproj
```diff
@@ -31,9 +31,8 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="chocolatey, Version=0.10.7.0, Culture=neutral, PublicKeyToken=79d02ea9cad655eb, processorArchitecture=MSIL">
-      <HintPath>..\packages\chocolatey.lib.0.10.7\lib\chocolatey.dll</HintPath>
-      <Private>True</Private>
+    <Reference Include="chocolatey, Version=0.10.8.0, Culture=neutral, PublicKeyToken=79d02ea9cad655eb, processorArchitecture=MSIL">
+      <HintPath>..\packages\chocolatey.lib.0.10.8\lib\chocolatey.dll</HintPath>
```
mwallner: not really a change request, but something I notice:
I don't like having path hierachies in nuget includes - maybe we should replace `..\packages` with something like `$(NuGetReferenceDir)` and then set the Property in a `Common.props` file?
- I'm working with 'global' nuget repository quite a bit - and therefore the `..\packages\..` paths just aren't correct (in my case)

ferventcoder: This is all managed with NuGet/Visual Studio, and these are hint paths. Anyway, when we upgrade it to Visual Studio 2017 project files, it will become much, much simpler.

Source/ChocolateyGui.Subprocess/ChocolateyService.cs
```diff
@@ -205,7 +202,6 @@ public Task<bool> IsElevated()
             var operationContext = OperationContext.Current;
             using (await Lock.ReadLockAsync())
             {
-                var originalConfig = Config.get_configuration_settings().deep_copy();
```
RichiCoder1: Does this most recent version fix the issue that necessitated this hack? Where config settings were getting carried across invokes?

ferventcoder: Yes, 0.10.8 has fixes for this

ferventcoder: Yes, 0.10.8 has fixes for this

PR #466, opened 2017-09-15
(GH-465) Use highestAvailable requested execution level instead of subprocess

### LINKED ISSUE #465: Use highestAvailable instead of subprocess
Move the subprocess code to a tag or branch for historical sake and move to a simpler model that requests admin if you are an admin, otherwise as a non-admin the GUI will just run under them.

ferventcoder:
When an administrator runs ChocolateyGUI, it should ask for elevation
to run. Then an administrator can do everything they need. When a
non-administrator runs ChocolateyGUI, which would be new functionality,
it should just run under their context. They will be able to do
whatever they can do under their user context. If that means nothing,
then the system will allow them to do that. If they can do things due
to where Chocolatey itself is installed or if they are using the
background service, then the system should behave appropriately.

Closes #465


Source/ChocolateyGui/Models/ChocolateySetting.cs
```diff
@@ -4,20 +4,14 @@
 // </copyright>
 // --------------------------------------------------------------------------------------------------------------------
 
-using System.Runtime.Serialization;
-
 namespace ChocolateyGui.Models
 {
-    [DataContract]
```
RichiCoder1: I won't miss these

Source/ChocolateyGui/Models/LogMessage.cs
```diff
@@ -1,23 +1,17 @@
...
 // </copyright>
 // --------------------------------------------------------------------------------------------------------------------
 
-using System.Runtime.Serialization;
-
 namespace ChocolateyGui.Models
 {
-    [DataContract]
-    public class StreamingLogMessage
+    public class LogMessage
```
RichiCoder1: Is this class strictly relevant anymore? With no more streaming, log messages should be able to go straight to the logger.

Source/ChocolateyGui/Services/ServiceCallbackHandler.cs
```diff
@@ -15,27 +15,27 @@ public ServiceCallbackHandler(IProgressService progressService)
             _progressService = progressService;
         }
 
-        public void LogMessage(StreamingLogMessage message)
+        public void LogMessage(LogMessage message)
```
RichiCoder1: Ahh, never mind.

Source/ChocolateyGui/ViewModels/RemoteSourceViewModel.cs
```diff
@@ -276,7 +278,7 @@ protected override void OnInitialize()
                             p.IsInstalled = true;
                         }
 
-                        Packages.Add(p);
+                        Packages.Add(Mapper.Map<IPackageViewModel>(p));
```
RichiCoder1: Huh, does this actually work? I didn't think Automapper likes interfaces.

Source/ChocolateyGui/Models/LogMessage.cs
ferventcoder: Probably not, but I didn't know what to do with the requests (and it was late)

Source/ChocolateyGui/ViewModels/RemoteSourceViewModel.cs
ferventcoder: I'm not sure, it would need to be tested. I bet you are correct.

Source/ChocolateyGui/Services/ServiceCallbackHandler.cs
ferventcoder: ☝️

Source/ChocolateyGui/Models/LogMessage.cs
ferventcoder: 👇

Source/ChocolateyGui/ViewModels/RemoteSourceViewModel.cs
ferventcoder: It was wrong. Pushed fix

Source/ChocolateyGui/ChocolateyExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+            new ConcurrentDictionary<string, CancellationTokenSource>(StringComparer.OrdinalIgnoreCase);
+
+        public static Task RunAsync(this GetChocolatey chocolatey)
+        {
+            return Task.Run(() => chocolatey.Run());
+        }
+
+        public static Task<ICollection<T>> ListAsync<T>(this GetChocolatey chocolatey)
+        {
+            return Task.Run(() => (ICollection<T>)chocolatey.List<T>().ToList());
```
mwallner: not really related to this change (as it's just been moved..) but there is a potentail crash/ uncaught exception:
image

ferventcoder: Oh joy...

Source/ChocolateyGui/ViewModels/RemoteSourceViewModel.cs
ferventcoder: Not quite fixed... it was using a package factory before.

Source/ChocolateyGui/ChocolateyExtensions.cs
mwallner: nevermind, all good
- just set my VisualStudio to go crazy on exceptions :-) (sorry for that ^^)

Source/ChocolateyGui/ViewModels/RemoteSourceViewModel.cs
ferventcoder: @chocolatey/chocolateygui can someone explain how this gets resolved by autofac? https://github.com/chocolatey/ChocolateyGUI/blob/c3ef83d27dd23fe2785435e4bebc7e291b454e44/Source/ChocolateyGui/Services/ChocolateyRemotePackageService.cs#L48

Source/ChocolateyGui/Services/ServiceCallbackHandler.cs
ferventcoder: It's likely the service callback handler could go away?

ferventcoder:
This definitely needs work, as logging isn't quite right, along with a few issues in loading some packages.

ferventcoder:
Just noticed this - when you edit a source, it stays down in the edit section and you have to leave and come back to edit a different source.

Source/ChocolateyGui/Services/ChocolateyService.cs
```diff
@@ -446,18 +434,17 @@ public Task<bool> IsElevated()
 
         public async Task<bool> RemoveSource(string id)
         {
-            var operationContext = OperationContext.Current;
             using (await Lock.WriteLockAsync())
             {
-                var chocoConfig = await GetConfigFile(operationContext);
+                var chocoConfig = await GetConfigFile();
```
RichiCoder1: Maybe you already reverted this, but I don't think these lines are stricktly necessary? Unless you had a reason to add them back?

ferventcoder:
I'm very happy with the changes here and I am ready to pull them in

gep13:
Okay, let's :shipit: 😸

PR #481, opened 2017-10-10
GH480: Persist current package details

gep13:
@RichiCoder1 okay, so this feels sort of clunky, but...

Thoughts?


gep13:
Could add some ignores to the Mapping Profile to ignore these values, but not sure if that is a wise move.

RichiCoder1:
Hrmm. Would be nice to figure out a way to pull those values at the same time. But getting package details is already pretty expensive....

gep13:
@RichiCoder1 I don't disagree.  Perhaps merge this and create a new issue to track improving this in a future release?

RichiCoder1:
The property I added in #482 will probably also need to be persisted.

gep13:
See related issue here: https://github.com/chocolatey/ChocolateyGUI/issues/485

PR #484, opened 2017-10-11
Match Gallery Styles

RichiCoder1:
Extract markdown default styles to their own CSS file, mirror Chocolatey Gallery site styles, and tweak foreground defaults.


ferventcoder:
Can you share images of what this looks like? I'm not a huge fan of what the package pages look like, so as long as you are steering clear there, then 👍

gep13:
@ferventcoder...

Before:

image

After:

image

PR #493, opened 2017-10-17
(GH-492) Switch to Markdig.wpf

### LINKED ISSUE #492: Switch markdown rendering to Markdig.Wpf
Currently we use CefSharp to render markdown as there wasn't any good alternatives. However, the excellent https://github.com/Kryptos-FR/markdig.wpf project has popped up which appears to do exactly what we need with high fidelity. It means we won't perfectly match the fidelity of the web package viewer, but it will greatly decrease the size and moderately decrease the load time of Gui.

RichiCoder1:
Closes #492;

There will be some incosistencies since this is a WPF flow document now instead of an HTML-rendered document, but I managed to keep the styling broadly this same. We can tweak it moving forward.

Related improvements: 
* Faster perf for rendering.
* No more markdown popping in after page load

With this we have officially dropped CefSharp. That should in and off itself increase perf (less dlls loaded), and removes a hard dependency we had on x86 and 4.5.2. Package size also has dropped by about 35mb (compressed).


ferventcoder:
Markdig!!

RichiCoder1:
Yup! I actually stumbled upon Markdig.wpf cuz I was planning to attempt writing my own and stumbled upon it. Markdig is awesome.

ferventcoder:
We've moved part of the website to Markdig and plan to move the entire thing the rest of the way there. The tool is fantastic!

PR #498, opened 2017-10-18
(build) Fix sign criteria for null paths

ferventcoder:
There was a thought that FileExists was robust enough to handle nulls
or empty strings and return false (because of course a null file
doesn't exist), but it appears that it simply does no validation and
allows the null value check against File.Exists, which errors on
missing paths. Check to be sure the value is empty prior to sending it
to the helper method.

### DIFF
--- a/setup.cake
+++ b/setup.cake
@@ -32,7 +32,7 @@ var CERT_ALGORITHM = EnvironmentVariable("CERT_ALGORITHM") ?? "Sha256";
 // So for now, we will just look to sign the executables prior to building MSI
 Task("SignExecutable")
     .WithCriteria(() => {
-       return FileExists(CERT_PATH);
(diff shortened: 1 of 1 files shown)


ferventcoder:
@gep13 @mwallner @RichiCoder1 this fixes the build issues.

gep13:
@ferventcoder so, I just did some digging into this as it was annoying me...

So, `FileExists` is handling a null being passed in (although, it is throwing an `ArgumentNullException`), the actual issue we were seeing was with the automatic conversion of a `string` to a `FilePath`.  This was resulting in a `ArgumentException` being thrown.

You can see this here:

https://github.com/cake-build/cake/blob/develop/src/Cake.Core/IO/Path.cs#L70

ferventcoder:
Thanks for the update! Would love it to just work without issue. The end user expectation for me would be that it would just return false because it communicates the same thing and it's not exceptional that there is no value.

Issue #517, opened 2017-10-27
Create a tile view for local/remote package views


gep13:
Which will compliment the existing list view that already exists.

Users should be able to switch easily between tile view and list view.

gep13:
The first part of this has now been added by @punker76:

![image](https://user-images.githubusercontent.com/1271146/32198043-6a385548-bdbe-11e7-8d20-87332b1343b7.png)

Additional PR's will handle follow up work for Local Package Source, etc.

PR #518, opened 2017-10-27
Tile View Implementation

gep13:
@RichiCoder1 @ferventcoder @punker76 @mwallner @pascalberger This is VERY much a Work In Progress, but I just wanted to put this out there for discussion.

What are your immediate thoughts on this?

User can switch between List and Tile view using buttons here:

image

Existing list view stays exactly the same as it is, but when switching to Tile, you see "something" like the following:

image

Interesting side effect of the InternetImage control that we are using, that when you first hit the new tile view, you see something like this:

image

So thoughts?  Good idea?  Bad idea?  What needs to be different?


ferventcoder:
Great start! I think we should have something for the placeholder images. Definitely like what I am seeing.

punker76:
@gep13 first thoughts

- caching images, cause switching to list and back loads it again
- ListView and ListViewItem style doesn't use MahApps one
- ToggleSwitch for the 2 views instead radio buttons

gep13:
@punker76 yes, image caching was something that had occurred to me as well.  Would need to extend the current control to look in the local cache before reaching out to the interwebs.

RichiCoder1:
> * caching images, cause switching to list and back loads it again

That's weird. If you're using the same control as on PackageView, it is caching. Maybe the the delay is reading from disk?

>* ToggleSwitch for the 2 views instead radio buttons

Agreed.

>* show also the version

As well as install/outdated status, preferably.

pascalberger:
@gep13 Sounds like a good idea to me. Also would like to see version and status (installed, outdated, pinned) in the view (maybe as overlays)

gep13:
@RichiCoder1 yip, it is the same control as on the package details view. Perhaps the issue is the number of requests at the same time.

punker76:
@RichiCoder1 @gep13 @pascalberger @ferventcoder Something for the eye...

new mockup 1

gep13:
@punker76 to me, this looks like exactly what we want 😄 

I think the word `New` might want to be another word, but not immediately sure on what it should be.  Part of me wants to put `Upgrade available` but that is obviously too much.

punker76:
@gep13 `out-of-date`

punker76:
@gep13 

new mockup 2

gep13:
Or perhaps the term outdated, which is what Chocolatey itself uses should be in there.

Either way, it is a small point, everything else looks :+1:

gep13:
@RichiCoder1 there is something up with the caching implementation.  Sometimes the Metadata isn't set, which means it goes back out to the internet to fetch the image again.  Sometimes it works, sometimes it doesn't 😢

punker76:
@gep13 @RichiCoder1 Caching images for InternetImage should be fixed now.

The InternetImage try to cache the images for 1 day and do this by adding an expired date as meta data. Setting this with the SetMetadata method for the file storage doesn't work. Requesting of this meta data returns a undefined DateTime.

But setting this directly to the meta data of the uploaded image stream works perfectly.

punker76:
@gep13 Intermediate result...

2017-10-28_22h22_52

mwallner:
to throw my fife cents in: I love what's going on, but I'd prefer a solution close to what Richard suggested - having more concrete borders around elements from the first place, without selecting anything .. (I think it'd be a cleaner interface)

punker76:
@mwallner I've thought of that, too...

new mockup 3

punker76:
@mwallner @gep13 

image

gep13:
@punker76 oooo, this is really starting to shape up, I like it!

One point... I think the name of the application should appear above the version number.

Also, while I think about it, while I don't dislike the current no icon symbol, I had thought about using a "washed out" version of the Chocolatey Icon, when there isn't a Icon Url defined.

What are your thoughts on that?

punker76:
@gep13 what do you mean with "washed out"?

Name above version...

image

gep13:
@punker76 said...
what do you mean with "washed out"?

i.e. use the Chocolatey Logo, but change it to a grey colour scheme and fade it out slightly.

I am sure that there is a technical term for what I am trying to describe, but I have no idea what it is 😄

punker76:
@gep13 @RichiCoder1 

Installed works now. But the `IsLatestVersion` property for this View says always `true`.

image

punker76:
@gep13 @RichiCoder1 @ferventcoder maybe I don't know enough about it, but it seems this could work...

```csharp
                    var installed = await _chocolateyPackageService.GetInstalledPackages();
                    var outdated = await _chocolateyPackageService.GetOutdatedPackages();

                    PageCount = (int)(((double)result.TotalCount / (double)PageSize) + 0.5);
                    Packages.Clear();
                    result.Packages.ToList().ForEach(p =>
                    {
                        if (installed.Any(package => package.Id == p.Id))
                        {
                            p.IsInstalled = true;
                        }
                        if (outdated.Any(package => package.Item1 == p.Id))
                        {
                            p.IsLatestVersion = false;
                        }

                        Packages.Add(Mapper.Map<IPackageViewModel>(p));
                    });
```

2017-10-29_12h11_37

punker76:
@gep13 @RichiCoder1 @ferventcoder @pascalberger I think I'm done for now, last one is the RadioButton style for the ListViewMode...

2017-10-29_14h47_48

2017-10-29_14h47_54

ferventcoder:
This looks amazing!! The only thing I might change - a watermarked default image when no image is available - that greyed out choco icon (maybe previously mentioned).

gep13:
@ferventcoder I am assuming in this case that you are ok with the official Chocolatey Logo being modified?  Or should this be an icon that makes it's way into the official Media Kit for Chocolatey?

ferventcoder:
This is an official Chocolatey client so :+1:

punker76:
@gep13 @ferventcoder 

2017-10-30_10h01_04

punker76:
@gep13 @ferventcoder also on all other places

2017-10-30_10h06_48

gep13:
@punker76 so, where are we with this?  Is there anything left that you are looking to add?  Or is this ready for review?

Should/could we add the same tile view to the Local Packages Tab as well?

Should/could we add an application level setting which specifies the default view to use within the application?

punker76:
> so, where are we with this? Is there anything left that you are looking to add? Or is this ready for review?

I'm done

> Should/could we add the same tile view to the Local Packages Tab as well?

That would be consistent. But we should do this maybe in a new PR.

> Should/could we add an application level setting which specifies the default view to use within the application?

👍

gep13:
@punker76 agree on new PR's for other pieces of work.  Let's get this part reviewed and pulled in, and we can discuss further.

I will try to take it for a proper spin tonight.

pascalberger:
I tried to install version from this build over my previous installation but after switching to tile view on the feed I received this error:

image

punker76:
@pascalberger Which prev version? The 0.15 or one prev of these PR? And can you try it again with the latest from this PR? Thx.

pascalberger:
Couldn't reproduce it with the latest build from this PR (but couldn't also reproduce it before either :) )

RichiCoder1:
Hmm. Looks like I didn't set a switch on Image Magick's svg conversion to use transparent backgrounds. Will need to enable that....

gep13:
@punker76 just had a play with this, and from what I can see, we can :shipit: and get some more feedback.

punker76:
@gep13 👏 so let me create the next one for the local view 😉 

choco_local_tileview

2017-10-30_22h52_35

gep13:
Relates to #517

PR #520, opened 2017-10-30
Local Package Tile View Implementation

punker76:
After implementing the tile view for the remote source it's more consequent to do this for the local package view too.

Changes
- Same tile view for local view
- Fix DataGrid sorting for Title and Authors

choco_local_tileview

2017-10-30_22h52_35

2017-10-31_00h03_29

Relates to #517

### DIFF
--- a/Source/ChocolateyGui/ChocolateyGui.csproj
+++ b/Source/ChocolateyGui/ChocolateyGui.csproj
@@ -204,6 +204,7 @@
     <Compile Include="Models\EnumToBoolConverter.cs" />
     <Compile Include="Models\LogLevel.cs" />
     <Compile Include="Models\LogMessage.cs" />
+    <Compile Include="Utilities\PackageAuthorsComparer.cs" />
     <Compile Include="Services\ChocolateyService.cs" />
     <Compile Include="Controls\InternetImage.xaml.cs">
       <DependentUpon>InternetImage.xaml</DependentUpon>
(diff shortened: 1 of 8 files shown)


RichiCoder1:
What's the bookmark symbol represent?

punker76:
@RichiCoder1 It's the `IsSideBySide` status.

punker76:
Sorting for Title and Authors fixed

choco_local_sorting

punker76:
@gep13 @RichiCoder1 I think I'm done for now. So there are only 2 things. Saving (and default) for the tile configuration. And second is, if we should also show a sorting ComboBox if the tile view is selected (where the user can sort by title, author, version, last version)?

RichiCoder1:
@punker76 Awesome work again :)

>  Saving (and default) for the tile configuration. 

At least for now I'd add a switch in settings like console output toggle.

> And second is, if we should also show a sorting ComboBox if the tile view is selected (where the user can sort by title, author, version, last version)?

Yup! I know that's going to be the one of the first requests we get after rolling this out

punker76:
@RichiCoder1 

> Yup! I know that's going to be the one of the first requests we get after rolling this out

ok, will do this in next days

gep13:
@punker76 epic work!

I have merged this into develop now, I think the outstanding items should be raised as separate issues, as the main part of #517 has been completed.

Issue #524, opened 2017-11-02
Remember tile/list view preferences


ferventcoder:
I noticed when I exit Chocolatey GUI and reopen it that it does not remember the last setting. My preference is the tile view by default and I think most people are also going to prefer that.

gep13:
@ferventcoder thanks for raising this issue, I have been meaning to add it.

PR #532, opened 2017-11-17
(GH-524) Added ability to default to Tile View

gep13:
- Left the list views as the default for starting application
- Choice is persisted and set on next start of the application

@ferventcoder @mwallner @RichiCoder1 @pascalberger @punker76 Thoughts on the following:

image


gep13:
Fixes #524

ferventcoder:
LGTM!

Issue #376, opened 2017-01-17
chocolateygui.portable package


ferventcoder:
This seems like a great add. Just a package that has the software in it with no dependencies.

RichiCoder1:
I don't see why this wouldn't be possible. We might have to tweak how logs are written so we don't write to ProgramData, but otherwise this seems like a fine idea to me.

pauby:
After internal discussion, we have agreed that offering a portable package is not something we are going to do.

PR #544, opened 2017-11-19
[WIP] (GH-376) Added new chocolatey packages

gep13:
- chocolateygui.portable which doesn't require a MSI installation
- chocolateygui.install which is a replacement for chocolateygui
- chocolateygui now takes a dependency on chocolateygui.install

### DIFF
--- a/nuspec/chocolatey/ChocolateyGUI.install.nuspec
+++ b/nuspec/chocolatey/ChocolateyGUI.install.nuspec
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package>
   <metadata>
-    <id>chocolateygui</id>
+    <id>chocolateygui.install</id>
     <version>$version$</version>
     <title>Chocolatey GUI</title>
     <authors>Chocolatey</authors>
@@ -32,7 +32,14 @@ All release notes for Chocolatey GUI can be found on the GitHub site - https://g
   </releaseNotes>
     <tags>chocolateygui chocolatey admin foss</tags>
     <dependencies>
-      <dependency id="chocolateygui.install" />
+      <dependency id="Chocolatey" version="[0.10.3, 0.11)" />
     </dependencies>
   </metadata>
+  <files>
+    <file src="chocolateyInstall.ps1" target="tools"/>
+    <file src="chocolateyUninstall.ps1" target="tools"/>
+    <file src="..\..\BuildArtifacts\ChocolateyGUI.msi" target="tools"/>
+    <file src="..\..\LICENSE" target="tools\LICENSE"/>
+    <file src="VERIFICATION.txt" target="tools"/>
+  </files>
 </package>
--- a/nuspec/chocolatey/ChocolateyGUI.install.nuspec
+++ b/nuspec/chocolatey/ChocolateyGUI.install.nuspec
@@ -10,7 +10,7 @@
     <projectSourceUrl>https://github.com/chocolatey/ChocolateyGUI</projectSourceUrl>
     <packageSourceUrl>https://github.com/chocolatey/ChocolateyGUI/tree/develop/ChocolateyPackage</packageSourceUrl>
     <iconUrl>https://cdn.rawgit.com/chocolatey/choco/8b6217061a4efe6543622e903996bb8eac012db2/docs/images/Choco_Icon.png</iconUrl>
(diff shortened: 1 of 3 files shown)


gep13:
@RichiCoder1 @ferventcoder here is a first pass at implementing this.

@RichiCoder1 with regard to the usage of the ProgramData folder for writing the log files, how were you planning on handling that?

nuspec/chocolatey/ChocolateyGUI.install.nuspec
```diff
@@ -0,0 +1,45 @@
...
+    <id>chocolateygui.install</id>
+    <version>$version$</version>
+    <title>Chocolatey GUI</title>
+    <authors>Chocolatey</authors>
+    <owners>Chocolatey</owners>
+    <projectUrl>https://github.com/chocolatey/ChocolateyGUI</projectUrl>
+    <projectSourceUrl>https://github.com/chocolatey/ChocolateyGUI</projectSourceUrl>
+    <packageSourceUrl>https://github.com/chocolatey/ChocolateyGUI/tree/develop/ChocolateyPackage</packageSourceUrl>
+    <iconUrl>https://cdn.rawgit.com/chocolatey/choco/8b6217061a4efe6543622e903996bb8eac012db2/docs/images/Choco_Icon.png</iconUrl>
+    <licenseUrl>https://raw.githubusercontent.com/chocolatey/ChocolateyGUI/develop/LICENSE</licenseUrl>
```
AdmiringWorm: just curious, why using the raw url?
Why not link to how it displayed on github, instead of using the download link?

nuspec/chocolatey/ChocolateyGUI.nuspec
```diff
@@ -24,22 +24,15 @@ Chocolatey GUI is a nice GUI on top of the Chocolatey command line tool.
...
+This package will only work correctly on Windows 7 SP1 through Windows 10 (1708) or Windows Server 2008 R2 SP1 through Windows Server 2016, and requires .NET Framework 4.5.2 at minimum.
   </description>
     <summary>A GUI for Chocolatey</summary>
   <releaseNotes>
 All release notes for Chocolatey GUI can be found on the GitHub site - https://github.com/chocolatey/ChocolateyGUI/releases
   </releaseNotes>
     <tags>chocolateygui chocolatey admin foss</tags>
     <dependencies>
-      <dependency id="Chocolatey" version="[0.10.3, 0.11)" />
+      <dependency id="chocolateygui.install" />
```
AdmiringWorm: I don't know if you remember the talk with Rob about using explicit version as a dependency a few months back, and why it was decided it should be used?

But basically, because then the user would be getting the version they specified to install on the meta package.

Isn't this possible to do with the following line?
```xml
<dependency id="chocolateygui.install" version="[$version$]" />
```
I've never tested this though, so I don't know if it works or not.

nuspec/chocolatey/ChocolateyGUI.install.nuspec
gep13: @AdmiringWorm good question.  Off the top of me head, I am not sure.  Would need to check the history.  Agree that this should be changed though.

nuspec/chocolatey/ChocolateyGUI.nuspec
gep13: Are you reading my todo list somehow?!?!  I had the same thought.  Have a note to check how this can best be done. 😄

RichiCoder1:
@gep13 There's nothing saying we have to write to ProgramData. We could write to the User's data directory. We could switch to doing this by default (breaking?) or we could look for some signal and write to the LocalAppData if that signal is set.

ferventcoder:
> We could write to the User's data directory. We could switch to doing this by default (breaking?) or we could look for some signal and write to the LocalAppData if that signal is set.

As long as this is not the default 👍 . By default it should be a machine location, even if it is portable.

gep13:
@RichiCoder1 I was thinking about checking to see where the exe is currently being ran from, and based on that, making an informed decision about where the log should be created.  Thoughts?

RichiCoder1:
Apologies for taking so long to respond! I know at least one other application that checks for a "portable" or equivalent signal and then writes to it's own folder instead. I'd be ok with that. Maybe we could add an extra file to portable packge to signal it being portable?

gep13:
@pauby based on the fact that we have closed the associated issue, it makes sense to go ahead and close this PR as well.

Issue #506, opened 2017-10-23
Latest Version not showing


daggmano:
The long awaited new version of Chocolatey GUI is here, allowing us to move the window during operations!  Fantastic, and the UI looks good.

Unfortunately, on my PC at least, the Latest Version column remains blank.  Even through I know there is an update awaiting me for at least one package, the column is blank and when I select 'Show Only Packages with Updates', the list is empty.

Let me know if you want any log files (and where they are).

RichiCoder1:
That is very odd. 
Can you produce the output of `choco outdated`?
Do you have any sources aside from the defaults?
Are there errors or warnings in your log files? If so, could you produce them? I'm available on gitter or twitter if they're in anyway sensitive.

daggmano:
Hmm, so there seems to be a path problem, perhaps?

    > choco outdated
    This is try 1/3. Retrying after 400 milliseconds.
     Error converted to warning:
     Access to the path 'C:\ProgramData\chocolatey\config\chocolatey.config.22644.update' is denied.
    This is try 2/3. Retrying after 600 milliseconds.
     Error converted to warning:
     Access to the path 'C:\ProgramData\chocolatey\config\chocolatey.config.22644.update' is denied.
    Maximum tries of 3 reached. Throwing error.
    Chocolatey v0.10.8
    Outdated Packages
     Output is package name | current version | available version | pinned?

    webstorm|2017.2.4|2017.2.5|false

    Chocolatey has determined 1 package(s) are outdated.

RichiCoder1:
No, the access denied is a weird quirk of Chocolatey of running from an unelevated prompted. For some reason it always tries to update the config. 

That's so weird. The output from `outdated` is correct :\. I'm not sure why it's not showing up.

daggmano:
Happy to wait a day or two to see if anyone else has any issues :)

daggmano:
@RichiCoder1  I uninstalled and reinstalled ChocolatelyGUI and all seems to be working now.  Is it correct that only apps where the latest version is different to the installed version will show version in the Lastet Version column?

RichiCoder1:
Yup, only apps with updates will show updates. Maybe there's some flaw in the logic that Gui uses that's accidentally filtering out webstorm. I'll see if I can reproduce. Thanks for the details!

PR #557, opened 2018-02-10
(GH-506): Corrected issue with package id

DamianMaslanka5:
In chocolatey when it gets webstorm package:
```IPackage availablePackage = packageManager.SourceRepository.FindPackage(packageName, version, config.Prerelease, allowUnlisted: false);```

It returns package with id ```WebStorm```. But when it gets local package:
```IPackage installedPackage = packageManager.LocalRepository.FindPackage(packageName);```
 it returns id ```webstorm```.

So there is probably a bug.
And in the fix chocolateyGUI takes packageName instead of id which seems to work.

### DIFF
--- a/Source/ChocolateyGui/Services/ChocolateyService.cs
+++ b/Source/ChocolateyGui/Services/ChocolateyService.cs
@@ -84,7 +84,7 @@ namespace ChocolateyGui.Services
                 var packages = await Task.Run(() => nugetService.upgrade_noop(chocoConfig, null));
                 var results = packages
                     .Where(p => !p.Value.Inconclusive)
-                    .Select(p => Tuple.Create(p.Value.Package.Id, p.Value.Package.Version.ToNormalizedString()))
+                    .Select(p => Tuple.Create(p.Value.Package.Id.ToLower(), p.Value.Package.Version.ToNormalizedString()))
                     .ToArray();
                 var parsed = results.Select(result => Tuple.Create(result.Item1, new SemanticVersion(result.Item2)));
 


gep13:
@DamianMaslanka5 thank you for submitting this fix.  Could you please provide some steps to reproduce the problem that you are seeing?

I just want to make sure that we are talking about the same thing.  Thanks

DamianMaslanka5:
@gep13 
Just install an old version of webstorm, for example 2017.3.3 and check your installed packages. It doesn't show latest version and you can't update the package.

gep13:
Relates to #506

gep13:
@RichiCoder1 can I get you to review this one?  Thanks!

RichiCoder1:
That is weird. Really weird. Id should be the same universally. Or I thought so. @ferventcoder ?

ferventcoder:
Id should not be case sensitive.

DamianMaslanka5:
@ferventcoder So after getting available package it should replace its id with id.to_lower() ?

gep13:
@DamianMaslanka5 yes, I think that makes sense.  I "think" what is going on here is that Chocolatey treats everything as lowercase, however, in this case, we are using the NuGetService to interrogate the packages and get the information about them.  As a result, we are seeing some package ids coming through without the lower case being applied.  Can you try with simply doing a switch to lower case and let me know if the problem is resolved?  If it is, if you can fix up this PR to reflect that, we can get it pulled in.

ferventcoder:
Following this up, case sensitivity shouldn't be an issue, but there is something somewhere that is likely applying it.

DamianMaslanka5:
@gep13 Updated.

gep13:
@DamianMaslanka5 thank you for taking the time to correct this issue.  I have merged this into the develop branch here: https://github.com/chocolatey/ChocolateyGUI/commit/bf0d7c129f7b661dd909bd08b52266a0f081449e

Issue #588, opened 2018-05-08
Search in all available repositories


mwallner:
This might come in handy for larger setups where "access-control" to packages is realized by having a multi-stream setup.
In such a szenario, the CLI `choco search` works just as expected, searching via the GUI becomes cumbersome. 

(searching for packages by typing in the search box should not be limited to the currently selected stream - maybe an option to "search in all streams" could be added to global settings)

mwallner:
will create an option to ignore source - entry points should be https://github.com/chocolatey/ChocolateyGUI/blob/develop/Source/ChocolateyGui/Services/ChocolateyService.cs#L167

PR #591, opened 2018-05-17
(GH-588) Add option to search packages in all available repositories

mwallner:
see #588

### DIFF
--- a/Source/ChocolateyGui/Models/AppConfiguration.cs
+++ b/Source/ChocolateyGui/Models/AppConfiguration.cs
@@ -17,5 +17,7 @@ namespace ChocolateyGui.Models
         public bool DefaultToTileViewForRemoteSource { get; set; }
 
         public bool UseDelayedSearch { get; set; }
+
+        public bool SearchInAllRepositories { get; set; }
     }
 }
--- a/Source/ChocolateyGui/Models/PackageSearchOptions.cs
+++ b/Source/ChocolateyGui/Models/PackageSearchOptions.cs
@@ -15,7 +15,8 @@ namespace ChocolateyGui.Models
             bool includePrerelease,
             bool includeAllVersions,
             bool matchWord,
-            string source)
+            string source,
+            bool searchInALlRepos)
             : this()
         {
             PageSize = pageSize;
@@ -25,6 +26,7 @@ namespace ChocolateyGui.Models
             IncludePrerelease = includePrerelease;
             MatchQuery = matchWord;
             Source = source;
+            SearchInAllRepos = searchInALlRepos;
         }
 
         public int CurrentPage { get; set;  }
@@ -32,6 +34,8 @@ namespace ChocolateyGui.Models
         public bool IncludeAllVersions { get; set; }
 
         public bool IncludePrerelease { get; set; }
+
+        public bool SearchInAllRepos { get; set; }
 
         public bool MatchQuery { get; set; }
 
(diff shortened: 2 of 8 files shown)


gep13:
@mwallner am I missing something here?  Or where are you actually telling it to search all repositories?  Perhaps we can talk about this one when we have a chat?

mwallner:
hey @gep13 - since I'm pretty busy currently here the explanation:
https://github.com/chocolatey/ChocolateyGUI/blob/1a34f8294914877b21c37e96d2af8d14a423c557/Source/ChocolateyGui/Services/ChocolateyService.cs#L168
(the logic is inverse - if "search in all repos" is active - the source to look in is _not_ set - otherwise it will be fixed to the currently active)

Source/ChocolateyGui/Models/AppConfiguration.cs
```diff
@@ -17,5 +17,7 @@ public class AppConfiguration
         public bool DefaultToTileViewForRemoteSource { get; set; }
 
         public bool UseDelayedSearch { get; set; }
+
+        public bool SearchInAllRepositories { get; set; }
```
gep13: Is this the right terms here?  Chocolatey GUI doesn't refer to repositories, but rather Sources.  Should this be `SearchInAllRepositories`?

Source/ChocolateyGui/Models/PackageSearchOptions.cs
```diff
@@ -15,7 +15,8 @@ public struct PackageSearchOptions
             bool includePrerelease,
             bool includeAllVersions,
             bool matchWord,
-            string source)
+            string source,
+            bool searchInALlRepos)
```
gep13: In addition to the other comment, this should be `searchInAllRepos`

Source/ChocolateyGui/Properties/Resources.resx
```diff
@@ -642,4 +642,7 @@ NOTE: Probably only necessary to change in RTL languages.</comment>
   <data name="PackageView_Version" xml:space="preserve">
     <value>Version</value>
   </data>
+  <data name="SettingsView_SearchInAllRepos" xml:space="preserve">
+    <value>Search in all Repos</value>
```
gep13: Depending on the outcome of the above, we will likely want to change this text as well.

gep13:
@mwallner I definitely prefer this being an opt-in option, so based on that, I have no objections.

Can you have a look at the comments that I made, and also update to the latest develop branch?  There are some changes that you will need to apply in the SettingsView to include a new description for the setting.  Let me know if you have a chance to do this.

Apologies in the time taken to get this reviewed.

Source/ChocolateyGui/Models/AppConfiguration.cs
punker76: I'm with @gep13 , so `SearchInAllSources` could be the name...

DamianMaslanka5:
@gep13 @mwallner Instead of enabling this option globally I would like to see `All` under `This PC`.
Right now when i go to private source i see my packages, and if I want to search for a specific package I can't when `Search in all Repos` is enabled. I need to go to options, disable it and then search again.

Probably there should be a way to hide `All`?

mwallner:
@DamianMaslanka5 - great idea!
@gep13 - I will see if I can pull this up .. :-)

mwallner:
@gep13 @DamianMaslanka5  - see GH-646
I'd be happy to close this PR if you're ok with the new one

gep13:
Closing in favour of #646

Issue #598, opened 2018-06-14
Separate Options/Sources from About screen


gep13:
Rather than having a single flyout with Options/Sources/About, there should be two flyouts.  One with Options/Sources and another with About.

This will make it easier to show/hide elements based on a users access to the application.

gep13:
What I am going to suggest is a change to this:

![image](https://user-images.githubusercontent.com/1271146/41435706-a06f8b70-7017-11e8-9595-5061ffbcae71.png)

Addition of an `about` link in the top right hand corner.

![image](https://user-images.githubusercontent.com/1271146/41435726-b411380e-7017-11e8-9f23-1767fb5f49da.png)

Changing the Settings screen to only have the `Options` and `Sources` tabs.

![image](https://user-images.githubusercontent.com/1271146/41435781-cd5afd9a-7017-11e8-8081-a4feda0ab3dd.png)

A new About View, that has three tabs, `History`, `Credits` and `Release Notes`.  Rather than trying to squeeze all the information into a single view as it was before.

![image](https://user-images.githubusercontent.com/1271146/41435868-006654aa-7018-11e8-8e20-e88f52ab2bdf.png)

gep13:
@punker76 @RichiCoder1 @pascalberger @mwallner what are your thoughts on the above?

punker76:
@gep13 👍  I think this looks better.

PR #599, opened 2018-06-15
(GH-598) Created new About View

gep13:
- To allow separation of the About and Settings
- This will allow easier control over how can access Settings in the future


gep13:
@punker76 yip, that is a good idea!  I will get that added.  Thanks!

gep13:
Relates to #598

PR #608, opened 2018-06-20
Upgrade MahApps.Metro and MahApps.Metro.IconPacks

### LINKED ISSUE #606: Don't allow navigation to settings or about screen while dialog is open
This will cause confusion to user, and shouldn't be allowed.

image

Wait for dialog to close, before allowing settings or about button to be clickable.

punker76:
This PR upgrades MahApps.Metro to v1.6.5 and MahApps.Metro.IconPacks to v2.2.0.

I also removed the obsolete MahApps.Metro.Resources package, which is no longer maintained or updated, so the future package is now always MahApps.Metro.IconPacks.

Relates to #606 Don't allow navigation to settings or about screen while dialog is open
Closes #606 

2018-06-20_10h23_42

2018-06-20_10h23_47


gep13:

Issue #618, opened 2018-09-12
Tile selection for Remote Sources is not being used


gep13:
Even though the setting for enabling Tile View by default has been selected, when changing to a remote source, the List View is enabled by default.

PR #619, opened 2018-09-13
(GH-618) Correct issue with ListViewMode default

gep13:
- Due to the fact that ViewModel, which hasn't been initialized, isn't regsitered
to subscribe to SettingChanged event, therefore it didn't "see" the change
- Move default ListViewMode setting into the initialize event to work around
this.


gep13:
Relates to #618

Issue #620, opened 2018-09-24
Unable to update source after making a change to another source


gep13:
Steps to reproduce:

* Edit one source (MyGet in screenshot below) and hit Save
* Click on another source, and notice that the source details section isn't populated with the source details
* Return to main screen and go back into Settings/Sources
* Click on source and source details are correctly populated

![image](https://user-images.githubusercontent.com/1271146/45944382-e5dfdb00-bfe1-11e8-90ac-fc48e1857679.png)

PR #621, opened 2018-10-06
(GH-620) Fix showing selected source after editing source.

DamianMaslanka5:
* Changed `SelectedItem` to `CurrentItem`
 `SelectedItem` is not getting updated after change to any source, I'm not sure why.
* Added `DraftSource` to prevent updating `SelectedSource` if save button was not clicked.
* Disabled multi select.

### DIFF
--- a/Source/ChocolateyGui/Models/ChocolateySource.cs
+++ b/Source/ChocolateyGui/Models/ChocolateySource.cs
@@ -12,7 +12,22 @@ namespace ChocolateyGui.Models
     {
         public ChocolateySource(ChocolateySource source)
         {
-            Update(source);
+            if (source == null)
+            {
+                throw new ArgumentNullException(nameof(source));
+            }
+
+            Id = source.Id;
+            Value = source.Value;
+            Disabled = source.Disabled;
+            UserName = source.UserName;
+            Password = source.Password;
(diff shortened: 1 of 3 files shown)


gep13:
Relates to #620

gep13:
@DamianMaslanka5 apologies for not getting to this sooner, I am only just seeing it now.

I have just taken this for a spin, and on the whole, it looks good, however, I am seeing one issue with this change.  If I do the following:

* select a source, make a change, and then save it
* click a different source
* click back onto the source that was changed
* click on the same different source

Then I get the following exception:

image

I haven't had a chance to dig into this yet, but I thought I would respond here in case you knew immediately what the problem is.

DamianMaslanka5:
@gep13 No problem. I fixed the issue and i found source of the bug. DataGrid is using GetHashCode and Equals but in ChocolateySource these methods are overriden and you probably know that if property is mutable it shouldn't be used in GetHashCode. If we don't modify the source directly it works well. If it looks good for you let me know and I will squash the commits, or you can do it.

gep13:
@DamianMaslanka5 I have just taken this for a spin, and it is now working perfectly.

Thank you again for taking the time to do this!

Issue #590, opened 2018-05-15
Next/Previous First/Last buttons remain disabled when using Simple.Server


gep13:
Link to Simple.Server:

https://github.com/chocolatey/simple-server

![image](https://user-images.githubusercontent.com/1271146/40058456-1c85a136-5849-11e8-9832-7bcbbddcde67.png)

Even though there are more than a page worth of packages that need to be displayed.

PR #623, opened 2018-10-20
(GH-590) Fix page count.

DamianMaslanka5:
Related to #590


gep13:
@DamianMaslanka5  You are on a roll!  Thanks for taking the time to fix these bugs!

gep13:
@DamianMaslanka5 not in a position to take this for a spin, but does this also fix the next/previous buttons as well?

DamianMaslanka5:
@gep13 On my machine next/previous button with simple server as source worked well without the fix I added 200+ packages so it showed 4 pages, I think problem described in #590 was that you had more than 50 packages but less than 100, page count was 1 instead of 2 and that's why next/previous buttons didn't work.

gep13:
@DamianMaslanka5 thanks for confirming.  That sounds like exactly what the problem was.  I will mark it as closed.  Thanks again!

PR #630, opened 2018-12-05
(GH-629) Separate app.manifest

### LINKED ISSUE #629: Separate the app.manifest from the exe
This will allow folks to make decisions on how best to set things when they run into interesting Windows issues. One issue being requested elevation when standard Windows users are members of groups that have some additional privilege elevation (including but in this case not the Administrators group), UAC will also pop up.  An admin using self-service that wants to remove this can simply make the adjustment to the app.manifest file to asInvoker.

https://github.com/chocolatey/ChocolateyGUI/blob/31f8544c08d3df52578136929ebafc430cba66cf/Source/ChocolateyGui/app.manifest#L7-L25

Related 
* https://github.com/chocolatey/ChocolateyGUI/issues/511#issuecomment-347241493 (just this comment)
* #342

ferventcoder:
By default highestAvailable is desired for UAC prompting behavior, but sometimes there are groups that provide some level of elevation (Network Configuration Operators Group) that also mean that UAC will
prompt. This is undesirable behavior in shops that might want to use self-service and have Chocolatey GUI run as the user that invoked without any elevation being requested. To allow this behavior, app.manifest can not be embedded, but must sit next to ChocolateyGUI.exe so that it can be changed for those situations that require it by those users.

Closes #629


ferventcoder:
This is in WIP because it appears there is some work to do to see that the ChocolateyGUI.exe.manifest gets included in the MSI

ferventcoder:
I took a cursory look but I don't know where I might need to change to be sure that ChocolateyGUI.exe.manifest gets included. The XSLT filters "ChocolateyGUI.exe" files out and then somewhere else the config and exe are put back in as named components. I think that the manifest needs the same treatment but I have not tracked down where that is just yet. @gep13 ?

gep13:
@ferventcoder I have just pushed a new commit onto your branch.

With this change in place, the MSI that is generated now includes the new manifest file, and on installation, it can be found here:

image

Next to the ChocolateyGUI.exe

ferventcoder:
Wow, whatever Github is doing, I can't rename this PR to remove "[WIP]".

Source/ChocolateyGui.Install/Product.wxs
```diff
@@ -92,6 +92,9 @@
       <Component Id="ChocolateyGui.exe.config" Guid="6C3AE56F-1AA3-4145-BFA9-345BBCC5C3D1">
         <File Id="ChocolateyGui.exe.config" Source="$(var.GuiTargetDir)/ChocolateyGui.exe.config" KeyPath="yes" Checksum="yes" />
       </Component>
+      <Component Id="ChocolateyGui.exe.manifest" Guid="AEA5D8AE-4274-444E-AEB5-988492C6919E">
+        <File Id="ChocolateyGui.exe.manifest" Source="$(var.GuiTargetDir)/ChocolateyGui.exe.manifest" KeyPath="yes" Checksum="yes" />
+      </Component>
```
ferventcoder: I figured it was something like that. Thanks for catching that!

ferventcoder:
And now it works. Deleting these two comments

ferventcoder:
And now it works. Deleting these two comments

ferventcoder:
We should add these notes in as well for folks changing this so they don't scratch their heads - https://github.com/chocolatey/choco/issues/1292#issuecomment-304068121

ferventcoder:
I added some notes to keep folks from being perplexed IF they do change this file. This is ready for review.

gep13:
@ferventcoder your changes have been merged, thanks for your contribution 👍

PR #634, opened 2019-01-05
(maint) Added Chinese language map

AdmiringWorm:
A new language was added on transifex which uses the full language code, as such
this pr adds a transifex language mapping so the language code would be in the expected format for .NET.
Without this change, the language would be mapped as 'Resources.zh_CN.resx' and possibly make the program crash

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,6 +1,6 @@
 [main]
 host = https://www.transifex.com
-lang_map = en_GB: en-GB, en_US: en-US
+lang_map = en_GB: en-GB, en_US: en-US, zh_CN: zh-CN
 
 [chocolatey-gui.resourcesresx]
 file_filter = Source/ChocolateyGui/Properties/Resources.<lang>.resx


gep13:
@mwallner I believe I am right in saying that the other languages that we have don't use a full language code, for example German uses `de`:

image

The latest language that has been added uses a full language code `zh_CN`:

image

And I believe that this is why the mapping is required, since .Net uses zh-CN (**NOTE:** the usage of a `-` rather than a `_`)

gep13:
@AdmiringWorm your changes have been merged, thanks for your contribution 👍

gep13:
@AdmiringWorm I also added a note about this in the documentation:

https://github.com/chocolatey/ChocolateyGUI/commit/274a277ddcedb0bcb5c22d15e011511c098c9b7f

AdmiringWorm:
@mwallner as @gep13 mentioned, this change isn't necessary to do when the culture is only two-part. And @gep13 is correct in his assessment of why the mapping is required.

mwallner:
cool, thanks for the clarification!

PR #646, opened 2019-01-26
(GH-588) Add option to show aggregated source view

mwallner:
This adds a setting option to show/hide an additional/virtual source that aggregates all sources.

(this aims to replace GH-591)

### DIFF
--- a/Source/ChocolateyGui/ChocolateyGui.csproj
+++ b/Source/ChocolateyGui/ChocolateyGui.csproj
@@ -232,6 +232,7 @@
     <Compile Include="ChocolateyExtensions.cs" />
     <Compile Include="Controls\MarkdownEx.cs" />
     <Compile Include="Enums\ListViewMode.cs" />
+    <Compile Include="Models\ChocolateyAggregatedSources.cs" />
     <Compile Include="Models\ChocolateyGUISetting.cs" />
     <Compile Include="Models\EnumToBoolConverter.cs" />
     <Compile Include="Models\LogLevel.cs" />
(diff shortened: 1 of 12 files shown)


mwallner:
hey @punker76 - could you have a quick look at this?
do you think there's an easy way of adding a separator between the local + aggregated and the "real sources" views?

I've got something like this in mind:
image

punker76:
@mwallner It's possible with a little trick to be MVVM like with an extra class which indicates as a separator and a StyleSelector.

I will make a commit with these changes, so all can see it and discuss them.

2019-02-18_12h20_01

gep13:
@mwallner the milestone doesn't need to be set on the PR, if the related issue has already been assigned to a milestone.

gep13:
Relates to #588

gep13:
@mwallner looks like you have used the wrong issue number in the commit messages, and also @punker76 has followed you 😄 

Can you update the commit messages to reference #588, rather than #646?

gep13:
@mwallner **NOTE:** @punker76 has just force pushed this branch, so in order to continue using your local copy of it, you will need to reset to the current head of this branch, otherwise, you will get inconsistencies, due to the re-writing of history.  Depending on your remotes, and what you have setup, you will need to do something _like_:

    git checkout all-sources-view
    git reset --hard origin/all-sources-view

Where origin is the remote to this repository.  If you have something else set up for this remote name, then you will need to change it accordingly.

Also, as a side note....

Going forward, can we **not** create branches directly in this repository, but rather in your own fork of the repository?  This reduces the amount of "churn" that needs to happen on the CI server.

gep13:
@mwallner @punker76 your changes have been merged, thanks you both for your contributions 👍

PR #658, opened 2019-04-19
(GH-501) Improve UI for Package view

### LINKED ISSUE #501: Package description should use all available client space
Package description doesn't seemes to be anchored correctly:

image

I would expect it to use full available height

punker76:
Sometimes the Release notes at the Package view seems like to be empty if there are only one line of text (e.g. #501).

To make this a little bit better I changed the UI to use Separator and increased Margins.

2019-04-19_23h13_15

2019-04-19_23h15_31

Closes #501


gep13:
@punker76 thanks again for doing this, really appreciate it!

Issue #665, opened 2019-04-26
Do not display password in source settings screen


gep13:
Once persisted into the chocolatey.config file, the password for a source (both user and certificate) are encrypted values, so there is no point in displaying them, as they will mean nothing to the end user.

Instead, when there is a password present, display a checkbox in the table view, and show a redacted textbox, when the source is selected for viewing/editing.

gep13:
For reference, we are talking about these sections in the sources settings screen:

![image](https://user-images.githubusercontent.com/1271146/56794149-76df9480-6805-11e9-90cc-98b0a26a0513.png)

gep13:
@punker76 in terms of UI changes, this is what I have so far...

![image](https://user-images.githubusercontent.com/1271146/56801633-d2198300-6815-11e9-8ca2-27d71882557d.png)

Thoughts?

gep13:
Just noticed that the `customer` TextBlock is in the wrong place.  I will need to fix this.

PR #666, opened 2019-04-26
(GH-665) Modifications to Sources Settings screen

gep13:
Don't allow direct viewing of Source Password (as this is encrypted and meaningless to end user)
Use ShowLoginAsync method in Mahapps to add/edit user and password

### DIFF
--- a/Source/ChocolateyGui/ViewModels/SettingsViewModel.cs
+++ b/Source/ChocolateyGui/ViewModels/SettingsViewModel.cs
@@ -9,6 +9,7 @@ using System.Collections.ObjectModel;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Threading.Tasks;
+using System.Windows;
 using System.Windows.Controls;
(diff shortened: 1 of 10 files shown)


Source/ChocolateyGui/ViewModels/SettingsViewModel.cs
```diff
@@ -328,6 +331,16 @@ public void Back()
             _eventAggregator.PublishOnUIThread(new SettingsGoBackMessage());
         }
 
+        public async void SetUserAndPassword()
+        {
+            var result = await _dialogCoordinator.ShowLoginAsync("Bob", "Bill", "Excellent", new LoginDialogSettings());
```
punker76: The first param should be the context (the registered vie model), like

```csharp
var result = await _dialogCoordinator.ShowLoginAsync(this, "Bill", "Excellent", new LoginDialogSettings { EnablePasswordPreview = true });
```

gep13:
Fixes #665

PR #667, opened 2019-04-26
(GH-633) Keep aspect ratio for Icons in Tile and Package view

### LINKED ISSUE #633: Icons: keep aspect ratio
Currently, icons are forced into a 1:1 ratio in the details section of a package. As most programs have an icon with a 1:1 ratio, that is fine.  

Sometimes, a software doesn't have an icon (because that software runs inside a host software hence there is no icon available) so you use a screenshot of the software from the developer's site. What chocolatey GUI does is turning [this image](https://www.voxengo.com/cdn/img/0/0lmtks3a-1.jpg) into this:  
grafik

So the image is quite distorted. It would be a nice improvement if ChocolateyGUI could keep the aspect ratio like it is done on the [webpage of the community server](https://chocolatey.org/packages/voxengo-stereo-touch/2.9).

punker76:
Keep aspect ratio for Icons in Tile and Package view.

Also make the download of the icons faster by using a memory stream instead a temporary file stream.

**Before**

2019-04-26_15h15_04

2019-04-26_15h15_39

**After**

2019-04-26_15h12_40

2019-04-26_15h12_57

Closes #633 

/cc @gep13 @mwallner @ferventcoder @pascalberger

### DIFF
--- a/Source/ChocolateyGui/Controls/InternetImage.xaml.cs
+++ b/Source/ChocolateyGui/Controls/InternetImage.xaml.cs
@@ -89,7 +89,7 @@ namespace ChocolateyGui.Controls
 
         private async Task<IBitmap> LoadImage(string url, float desiredWidth, DateTime absoluteExpiration)
         {
-            var imageStream = await DownloadUrl(url, desiredWidth, absoluteExpiration);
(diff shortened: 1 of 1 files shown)


gep13:

Issue #525, opened 2017-11-02
On Load of any source, outdated is running (logging) every time


ferventcoder:
Not sure if anyone else is seeing this as well. Outdated appears to be in the logs when switching between sources.

RichiCoder1:
That sounds about right, especially now that we check for outdated packages in remote views too. Will need to do a better job of caching results.

ferventcoder:
At least cache for a few minutes.

gep13:
In implementing this, a few changes have been made.

Settings screen has been broken out into two separate tabs, one for Chocolatey GUI and one for Chocolatey:

![image](https://user-images.githubusercontent.com/1271146/57125586-e4e60780-6d81-11e9-958f-0ab21ca8d062.png)

![image](https://user-images.githubusercontent.com/1271146/57125602-f3342380-6d81-11e9-90c6-47313ba66603.png)

Now, functionality in terms of how Features and Settings are handled are common between Chocolatey GUI and Chocolatey.

CLI functionality has been extended to include the ability to set Config values through the CLI:

![image](https://user-images.githubusercontent.com/1271146/57125680-2bd3fd00-6d82-11e9-829a-9995f9caebc1.png)

![image](https://user-images.githubusercontent.com/1271146/57125701-3b534600-6d82-11e9-83f9-87d48e412ebd.png)

![image](https://user-images.githubusercontent.com/1271146/57125717-49a16200-6d82-11e9-80a3-02e2bc7d87c2.png)

![image](https://user-images.githubusercontent.com/1271146/57125737-57ef7e00-6d82-11e9-9191-f33998ede564.png)

All of the above output is localized, and once merged can be translated via Transifex, so any help in this area (@mwallner @pascalberger @AdmiringWorm @punker76) would be gratefully appreciated.

gep13:
Here are some screenshots of the above working in Dutch :-)  @ferventcoder thought you might be interested in seeing these.

![image](https://user-images.githubusercontent.com/1271146/57174960-394ebd00-6e3e-11e9-9306-25a63e80bda7.png)

![image](https://user-images.githubusercontent.com/1271146/57174966-45d31580-6e3e-11e9-8bcc-b91b32b45cd5.png)

![image](https://user-images.githubusercontent.com/1271146/57174971-508daa80-6e3e-11e9-9b79-9a4d7eccb899.png)

Thanks @mkevenaar for creating the screenshots!

gep13:
[...]

PR #669, opened 2019-05-03
(GH-525) Add ability to cache information about outdated packages

gep13:
This includes a new configuration value, which default to 60, which is the amount of time in minutes for the cache file to be used.


gep13:
Fixes #525

Issue #680, opened 2019-06-11
Remove tilting chocolatey logo


gep13:
Based on new recommendations for usage of logo, and also to match the changes that are coming to the Chocolatey website, the Chocolatey Logo in the top left corner of Chocolatey GUI should no longer move.

gep13:
![image](https://user-images.githubusercontent.com/1271146/59261904-28ac1680-8c36-11e9-9a1f-8eefa9b5c016.png)

gep13:
@ferventcoder @designxsteph this is my first stab at implementing this.  Thoughts?

designxsteph:
@gep13 Looks good to me!

PR #681, opened 2019-06-11
(GH-680) Remove tilting Chocolatey Logo

gep13:
So that Chocolatey GUI more closely matches the look and feel of new
chocolatey.org redesign, and also meets with the new icon usage
guidelines.


gep13:
Fixes #680

PR #692, opened 2019-07-30
(build) Disable documentation of build script tasks

pascalberger:
Disable generation of build script tasks documentation

### DIFF
--- a/recipe.cake
+++ b/recipe.cake
@@ -31,7 +31,8 @@ BuildParameters.SetParameters(context: Context,
                             shouldPublishChocolatey: false,
                             shouldPublishNuGet: false,
                             shouldPublishGitHub: false,
-                            shouldExecuteGitLink: false);
+                            shouldExecuteGitLink: false,
+                            shouldDeployGraphDocumentation: false);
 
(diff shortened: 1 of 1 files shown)


gep13:
@pascalberger I am not too concerned either way, but was there a compelling reason for turning this off?

pascalberger:
@gep13 I think the question is what's the target audience of this documentation? End-users or developers?

End-users would not be interested in the build scripts available to building the application. Having a "Tasks" menu item that prominent, which also doesn't explain what kind of tasks it documents, is probably quite confusing for them.

For developers having a list of tasks provided by the underlying build script recipe is probably also not that much of an intereset as it doesn't show the actual build process used for this repo (e.g. some tasks are disabled by configuration but still shown here). If we want to have documentation about the build process for developers I would find it much more helpful to have a chapter somewhere mentioning that it uses Cake.Recipe with a link to the Cake.Recipe documentation where the individual tasks already are documented, and additional documentation about repo specific configuration and additional tasks. Or topic specific documentation, like done for the [Localization Howto](https://chocolatey.github.io/ChocolateyGUI/docs/localization-howto).

PR #715, opened 2019-11-15
(GH-688) Inline context search

### LINKED ISSUE #688: Inline context sensitive KB input search ("jump")
Instead of using Search:...
(A search enhancement.)

It would be nice if Chocolatey GUI would accept a typed letter from the keyboard and then jump to the first package name beginning with that letter and highlight it. As well as two characters (etc.) to jump to the first package matching the KB input. Ditto if sorted by Authors.

daveMueller:
closes #688

### DIFF
--- a/Source/ChocolateyGui.Common.Windows/Views/LocalSourceView.xaml
+++ b/Source/ChocolateyGui.Common.Windows/Views/LocalSourceView.xaml
@@ -247,7 +247,9 @@
                                 <DataGrid ItemsSource="{Binding PackageSource, Mode=OneWay}"
                                           AutoGenerateColumns="False"
                                           IsReadOnly="True"
-                                          SelectionMode="Single">
+                                          SelectionMode="Single"
+                                          IsTextSearchEnabled="True"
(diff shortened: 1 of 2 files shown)


daveMueller:
I this PR Chocolatey GUI jumps to the first package title beginning with the typed letter/letters and highlights it for the ```DataGrid``` (local packages) or the ```ListView``` (remote packages) that is currently displayed. That means e.g. for the remote sources, when typing the letter 'e' it doesn't search for all remote packages starting with 'e'. It only jumps to the first package starting with 'e' in the currently rendered ```ListView```.

Also currently it is only searching for the title and not the authors. They way I see it is that the current search field of the Chocolatey GUI also doesn't search for authors. Maybe I'm wrong with that.
Nevertheless I can't see a valid trigger to switch the search to authors. Switching the search to authors when the grid is sorted by authors isn't really transparent for the user. This would only lead to a strange UX.

gep13:
@daveMueller your changes have been merged, thanks for your contribution 👍

PR #724, opened 2020-01-06
(GH-711) prelease label should not be red

### LINKED ISSUE #711: prelease label should not be red
Currently the prerelease label is red:

image

I would suggest to change this to a yellow/orange colour. 

IMO a red color shows that it's bad. I think a prerelease is more a warning.

mkevenaar:
When a label is red, something wrong is implied, switching to DarkOrange
Implies a warning.

image

Merging this fixes #711


gep13:
@mkevenaar your changes have been merged, thanks for your contribution 👍

PR #729, opened 2020-02-02
Add Czech language map

pcpepik:
Adds a mapping for the Czech language code
so it would be in the correct format for .NET


pcpepik:
There's probably no need for Czech language to use a full language code, but oh well, it uses one, thanks to my request on transifex... This PR is to fix possible implications mentioned in the documentation.

gep13:
@pcpepik your changes have been merged, thanks for your contribution 👍

Issue #706, opened 2019-10-24
Settings to prevent preloading on remote views


iammael:
Hello!

I'd suggest to add this toggler :

![cxiVhXElNo](https://user-images.githubusercontent.com/7931135/67447066-fa423d00-f612-11e9-982e-1e78fb28b8c5.png)

It makes the page load a little bit faster. I have the code done for it and could submit a PR on approval (and yes I added a dot at the end of the description in the code).

gep13:
@iammael This looks like an interesting addition.  If you can send a PR, we can review the code.  Thanks for your interest in this project.

iammael:
Thank you for your answer. I'll send it when I get back home in 3 days!

opusforlife2:
@iammael Have you taken a look at this again?

iammael:
Hi!

I'm really sorry, I didn't have time to log on Github for a while.

Is this still something you're interested in ?

opusforlife2:
I definitely am.

@gep13 Could you confirm that this isn't already done/being done so that the effort doesn't get duplicated?

gep13:
@opusforlife2 no work has/is currently been done on implementing this issue.

opusforlife2:
Wow! That was fast! Thanks!

opusforlife2:
@gep13 What's the reason behind the slow release cadence? Is there a reason why there aren't smaller releases with more frequent bug fixes and feature additions?

gep13:
There are lots of reason why the 0.17.0 release took so long, but going forward, we are not planning on making the next release wait so long.  Having said that, I don't know when the next release will be.

If you want to take the pre-release of Chocolatey GUI for a spin, you can definitely do that.  The instructions are on the readme of this repository.

PR #740, opened 2020-03-29
(GH-706) Add a settings to prevent searchs preloading

iammael:
Add a toggle in the settings view to prevent a blank search when viewing
the list of available packages.

https://github.com/chocolatey/ChocolateyGUI/issues/706


gep13:
@iammael thank you very much for this addition!  I think we will need a follow up change to provide some information about why no packages are being displayed, and also to add an empty itemtemplate to the ListView as well, but that can be covered in another issue.

PR #751, opened 2020-04-05
(GH-739) Improving image quality and fixing issues with "ico" files

### LINKED ISSUE #739: Improve how application icons are rendered within Chocolatey GUI
Hi there,

i just noticed that "ico" files aren't displayed anymore.

This can be fixed by telling `MagickImage`, via `MagickReadSettings`, that it's an "Ico".
When doing so the image quality is very bad.
In case of "ico" the current image loading loads the first frame present in the "ico" which means it's almost always 16x16.

I made some changes to improve the image quality on my machine.

Before:
image

After:
image

That's the part for "ico" files.

"SVG" files don't seem to be rendered/loaded correctly too.
Currently this is what i get for Firefox:
image

Would you be interested in a PR changing the decoding of images to different kinds of decoders for different image formats?
My PR would only affect "ico" currently, as i am not familiar with `MagickImage` and how to fix the wrong "SVG" rendering.

batzen:
This fixes #739

### DIFF
--- a/Source/ChocolateyGui.Common.Windows/Controls/InternetImage.xaml.cs
+++ b/Source/ChocolateyGui.Common.Windows/Controls/InternetImage.xaml.cs
@@ -87,6 +87,34 @@ namespace ChocolateyGui.Common.Windows.Controls
             imageStream.Position = 0;
         }
 
+        private static IMagickImage ExtractImage(MagickImageCollection imageCollection, Size desiredSize)
+        {
+            var imagesOrderedBySize = imageCollection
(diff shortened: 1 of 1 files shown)


gep13:
@batzen thanks for doing this!  I will take a proper look at this later, and get it merged in.  I added a couple of commits into your branch to fix the build error that was happening on AppVeyor.

gep13:
@batzen thank you for taking the time to dig into this, and for submitting this PR, really appreciate it!

Issue #742, opened 2020-03-30
Explain why there are no packages in remote views


gep13:
As a follow up to this [issue](https://github.com/chocolatey/ChocolateyGUI/issues/706) it would be good to provide some text when 

- [x] first going to a remote source, to explain why there are no packages displayed
- [x] when searching a remote source and no packages are found
- [x] when searching a local source and no packages are found

gep13:
Part of this has been completed, and merged into the develop branch.  Now, when initially opening a remote source and the Prevent Preload feature is enabled, a user will see the following:

![image](https://user-images.githubusercontent.com/1271146/78422962-f8b58000-765a-11ea-8212-85d7da620a0b.png)

gep13:
@punker76 has now added a message when there are no packages found when searching:

https://user-images.githubusercontent.com/658431/78460887-10c1e980-76c5-11ea-8fd7-9d6a9a1ea5ec.gif

gep13:
@punker76 I have re-opened this issue, as I believe we will need to add the same empty item data template to the local source, when searching for packages there.  I would like to take a stab at this, as I "think" I should be able to use what you have implemented in the remote source as a guide for what to do in the remote source.

gep13:
I am happy to say that @punker76 was able to complete the remaining work, which added a similar empty item template to the local source view, which you can see here:

![choco_search_result_local2](https://user-images.githubusercontent.com/658431/78645241-81c8f300-78b7-11ea-9dc6-735b77df1bdb.gif)

Which means that this can now be closed out.

gep13:
I am going to remove this issue from the 0.18.0 milestone, as the main scope of work was done in this issue: https://github.com/chocolatey/ChocolateyGUI/issues/706 and therefore already covered in the generated release notes.

PR #752, opened 2020-04-06
(GH-742) Add empty item template to local source

gep13:


gep13:
@punker76 can you please tell me why the empty item template isn't working on the local source?

I have tried to copy/paste your code, but there must be a bit more to it, and hoping that you can help :-)

punker76:
@gep13 The empty results info should now work for tile and datagrid view.

choco_search_result_local

punker76:
@gep13 I improved a little bit more here. The trigger for the empty template looks now also to the IsLoaded property to prevent showing it if the user refresh packages. The other one is the 250ms delay on search text box, so that the search is done after this and not directly on every user input.

punker76:
choco_search_result_local2

gep13:
@punker76 quick question... Does this:

> The other one is the 250ms delay on search text box, so that the search is done after this and not directly on every user input.

Take into account the UseDelayedSearch feature?

image

punker76:
@gep13 The UseDelayedSearch is currently only for the remote service, because of long running task. IMO the local view doesn’t need that, because the search goes directly to the CollectionView and is faster.

gep13:
@punker76 that is a great point!  I think we need to update the description to reflect that this only applies to the remote sources, but I can do that separately.

PR #760, opened 2020-04-10
Update Automapper to v7.0.1

punker76:
With this version I can fix an exception which I got at the `ChocolateyService` with the old one.

```csharp
        private static Package GetMappedPackage(GetChocolatey choco, PackageResult package, IMapper mapper, bool forceInstalled = false)
        {
            var mappedPackage = package == null ? null : mapper.Map<Package>(package.Package);
            if (mappedPackage != null)
```

System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.

v7.0.1 is also the latest possible version for .Net v4.5.2


gep13:

PR #776, opened 2020-05-20
(GH-775) Fix search at remote source

### LINKED ISSUE #775: Could not search on remote source after changed from GH-771
This happens because of the wrong parameters now on LoadPackages. Better changing this to SearchForPackages and CanSearchForPackages to prevent such a mistake in the future.

image

punker76:
Use different method names to prevent such a mistake in the future.

image

Closes #775


gep13:
@punker76 thanks for fixing this up!

Issue #602, opened 2018-06-18
Global Chocolatey GUI settings


gep13:
At the minute, the Chocolatey GUI settings:

![image](https://user-images.githubusercontent.com/1271146/41556177-dd26eef6-7331-11e8-86e4-e7416b33ef24.png)

are configured on a person user basis.  It should be possible to allow settings these options globally, with the ability for users to configure these to their own needs, if required.  They will help with administrative deployments of Chocolatey GUI.

This should include the ability to roll back setting to the globally configured defaults.

Also, for an administrative user, when making a change, it should be possible to specify whether the change is to the global setting, or to the local user setting.

gep13:
Following a discussion with @ferventcoder it makes sense that these global configuration options should be added to the main chocolatey.config file.  This is already starting to house configuration for other parts of the Chocolatey ecosystem, like CCM and Chocolatey Agent, so it makes sense that this should also house configuration data for Chocolatey GUI.

gep13:
This has now been implemented, and is available in the latest package from the Chocolatey MyGet feed (see the readme for more information) and it will ship as part of the 0.18.0 release of Chocolatey GUI.

Chocolatey GUI has always had a data.db file in the users appdata folder:

![image](https://user-images.githubusercontent.com/1271146/105696197-e57de480-5efa-11eb-8006-2b5046f90e07.png)

which housed all the user specific configuration for Chocolatey GUI.

With this change, there is now also another data.db file, located in the common appdata folder

![image](https://user-images.githubusercontent.com/1271146/105696113-ca12d980-5efa-11eb-974c-be0c6c6dbda3.png)

which will house all the global defaults for the Chocolatey GUI configuration.

Opening Chocolatey GUI and using the Settings screen will always update/edit the configuration within the user specific data.db file:

![image](https://user-images.githubusercontent.com/1271146/105696405-2f66ca80-5efb-11eb-9bfb-f7c65322a888.png)

However, using the `chocolateyguicli` command, it is now possible to list out, and modify both the global configuration and the effective user specific configuration.

For example, running:

```
chocolateyguicli feature list
```

will list the current effective configuration for the currently logged in user:

![image](https://user-images.githubusercontent.com/1271146/105696682-8f5d7100-5efb-11eb-82bf-b01470ea7cde.png)

and running:

```
chocolateyguicli feature list --global
```

Will show the global configuration that is in place for all users:

[...]

PR #781, opened 2020-06-15
(GH-602) Initial attempt at getting things working

gep13:
Very early days, but things seem to be working....

Relates to #602

### DIFF
--- a/Source/ChocolateyGui.Common/Services/ConfigService.cs
+++ b/Source/ChocolateyGui.Common/Services/ConfigService.cs
@@ -216,20 +216,17 @@ namespace ChocolateyGui.Common.Services
 
         public void ListSettings(ChocolateyGuiConfiguration configuration)
         {
-            Logger.Warning(Resources.Command_SettingsTitle);
-            Logger.Information(string.Empty);
-
             foreach (var setting in GetSettings(configuration.Global))
             {
-                Logger.Information("{0} = {1} | {2}".format_with(setting.Key, setting.Value, setting.Description));
+                if (configuration.RegularOutput)
+                {
+                    Logger.Information("{0} = {1} - {2}".format_with(setting.Key, setting.Value, setting.Description));
+                }
+                else
+                {
+                    Logger.Information("{0}|{1}|{2}".format_with(setting.Key, setting.Value, setting.Description));
+                }
             }
-
-            Logger.Information(string.Empty);
-            Logger.Warning(Resources.Command_FeaturesTitle);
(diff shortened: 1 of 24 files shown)


gep13:
@AdmiringWorm @steviecoaster @mwallner @mkevenaar @punker76 @adilio this is the output from the stream tonight.  Could you all have a look when you have a minute.  Things seem to be working, but haven't really tested it out fully yet.

Source/ChocolateyGui.Common/Services/ConfigService.cs
```diff
@@ -147,13 +222,14 @@ public void EnableFeature(ChocolateyGuiConfiguration configuration)
 
         public void DisableFeature(ChocolateyGuiConfiguration configuration)
         {
+            var chosenAppConfiguration = configuration.FeatureCommand.Global ? _globalAppConfiguration : _userAppConfiguration;
             var featureProperty = GetProperty(configuration.FeatureCommand.Name, true);
-            var featureValue = GetPropertyValue<bool>(_appConfiguration, featureProperty);
+            var featureValue = GetPropertyValue<bool?>(chosenAppConfiguration, featureProperty);
```
AdmiringWorm, line 227: I have been wondering, maybe an alternative aproach here could be taken instead of using a nullable bool?

Like for instance passing having an overload that allows a default value to return if the actual value is null.

mwallner: I'm not a C# person, but from a PL perspective I'd rather not "re-introduce" nullable types whenever possible.

In general, I'd prefer a more simplistic approach such as only storing default/global values in the global db and only have user-specific values in the user db.
* if you use the `--global` switch, set the value in global db and remove the key/value pair from the users db ..
* if the user changes a setting, only set it in the user db.
whenever accessing a config value, use some proxy method that first checks if the key exists in the user database, and return either that or the global attribute (if no user-specific setting is present)

That said - again, I'm really not a C# person - just don't like nullable types :-D

gep13: @AdmiringWorm @mwallner sorry for not replying to either of you sooner about this.  I have just come back around to look at this again, and I am going to proceed with the current implementation.  I believe that this provides the functionality that is required, and covers all the edge cases regarding whether a global default value has been set or not, without over complicating the matter.

I have just taken this for a spin again this morning, alongside the Chocolatey GUI Licensed Extension, and I am happy to say that it is working as expected, so for now, I am going to get this merged.

PR #784, opened 2020-07-10
Removing all the SolutionInfo.cs files from the solution.

bhargava-raju:
These files no longer exists in solution folder so removing all the references in project folders


gep13:
@bhargava-raju thank you for taking the time to submit this, however, I won’t be able to accept this PR. You have to first run the build outside of Visual Studio using either the build.ps1 or the build.sh file. This will generate the missing file, and then everything should work.

PR #812, opened 2021-02-12
Update Svg.Skia and some SkiaSharp dependencies

punker76:
Also fix possible NullPointerException in ExtractImageFromStream. If a Svg could not be loaded the Picture property is null.


gep13 (approved):
LGTM!

PR #817, opened 2021-03-05
Feature/startup fix

Janek91:
Someone deleted SolutionInfo.cs file from repo and that caused solution refused building itself. Also, there was missing "Config" directory creation for DB on startup - that caused app crash on startup.
By the way - I've converted packages.config to PackageReference to speed up compilation and make it easier to manage NuGet packages.


gep13:
@Janek91 thank you for your interest in helping with Chocolatey GUI.  Before going too much further though, can I ask if you have read the contributing guide?

https://github.com/chocolatey/chocolateygui/blob/develop/CONTRIBUTING.md

The SolutionInfo.cs file is intentionally left out of source control, as this file changes on each build.  In order to generate it for the first time, you would need to run the full build once, before attempting to build the solution in Visual Studio.  This is mentioned in the [readme](https://github.com/chocolatey/ChocolateyGUI#building).

There are a few moving parts in terms of how the build process works for Chocolatey GUI, includes searching in the `packages` folder that is created when restoring packages from NuGet.  The change that you are suggesting here breaks this, as can be seen in the failed AppVeyor build.  This change, if moved forward, would need to be discussed in a separate issue, and updated in a different PR.

Regarding that Config folder.  That is an interesting one.  That folder is created as part of the WiX installation script:

https://github.com/chocolatey/ChocolateyGUI/blob/46c4a2ae5ecac4e20c6c8d0fccffa2411aeccce9/Source/ChocolateyGui.Install/Product.wxs#L74-L81

But perhaps something does need to be done here to aid with testing of the project during development.  I think it would make sense to discuss this further as an issue, before any other action is taken.

gep13:
@Janek91 we haven't heard back from you about this PR so for now, I am going to go ahead and close it out, but feel free to respond here and we can discuss further.

Issue #757, opened 2020-04-08
Provide ability to control which Chocolatey operations are allowed


gep13:
At the minute, all Chocolatey operations (install, upgrade, pin, uninstall, etc) within Chocolatey GUI are always allowed to be executed.  However, it would be good to control which of these operations can be executed.

By default, this feature should have all operations accessible, but over time, different conditions may be added that allow turning off certain operations.

PR #819, opened 2021-03-06
(#757) Allow buttons to be enabled/disabled

gep13:
The decision was taken to allow a button to be enabled/disabled based on
current Chocolatey configuration (mainly when running in background
service mode). To allow for this, whilst still allowing a button to be
enabled/disabled based on application logic, a new behavior was
introduced to encapsulate this logic.

This will allow the Chocolatey GUI Licensed Extension to override the
AllowedCommandsService, to control which operations should be enabled
and disabled, based on the background service configuration.

Relates to #757

### DIFF
--- a/Source/ChocolateyGui.Common.Windows/Resources/Controls.xaml
+++ b/Source/ChocolateyGui.Common.Windows/Resources/Controls.xaml
@@ -10,6 +10,8 @@
                     xmlns:iconPacks="http://metro.mahapps.com/winfx/xaml/iconpacks"
                     xmlns:mah="http://metro.mahapps.com/winfx/xaml/controls"
                     xmlns:theming="clr-namespace:ChocolateyGui.Common.Windows.Theming"
+                    xmlns:b="http://schemas.microsoft.com/xaml/behaviors"
+                    xmlns:utilities="clr-namespace:ChocolateyGui.Common.Windows.Utilities"
                     mc:Ignorable="d">
 
     <ResourceDictionary.MergedDictionaries>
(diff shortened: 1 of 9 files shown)


Source/ChocolateyGui.Common.Windows/Resources/Controls.xaml
```diff
@@ -451,31 +453,73 @@
         <MenuItem Header="{x:Static properties:Resources.Controls_PackagesContextMenuPin}"
                   Icon="{iconPacks:Modern Kind=Pin}"
                   Command="{commands:DataContextCommandAdapter Pin}"
-                  Visibility="{Binding CanPin, Converter={StaticResource BooleanToVisibility}}" />
+                  Visibility="{Binding CanPin, Converter={StaticResource BooleanToVisibility}}">
+            <b:Interaction.Behaviors>
+                <utilities:ToolTipBehavior EnabledToolTip="{x:Static properties:Resources.Application_OperationPin}"
+                                           DisabledToolTip=""
```
punker76, line 459: Set this to an empty value is not necessary.

Source/ChocolateyGui.Common.Windows/ViewModels/Items/PackageViewModel.cs
```diff
@@ -148,17 +148,29 @@ public string[] Authors
             set { SetPropertyValue(ref _authors, value); }
         }
 
-        public bool CanInstall => _allowedCommandsService.IsInstallCommandAllowed && !IsInstalled;
+        public bool CanInstall => !IsInstalled;
```
punker76, line 151: The `Can...` properties needs the allowed check too. Without that the buttons/menus are allowed to use.

punker76 (commented):
I think the `Can...` properties still needs also the allowed check. Without that you will see the disabled feature tooltip but can use the button/menu.

Source/ChocolateyGui.Common.Windows/ViewModels/Items/PackageViewModel.cs
gep13: The `CanInstall` boolean is tied to the Visibility property of the button.  So I am not sure if we want to add the AllowedCommands property to this.  Before this PR, I was showing/hiding the buttons based on whether the operation was allowed and if the operation currently made sense.  With this change, the button is now shown/hidden based on whether the operation makes sense (i.e. the install button is shown, if the package isn't currently installed), and then using the new behavior, it will either be enabled/disabled, based on whether the operation is allowed based on the current Chocolatey configuration.

Does this make sense?  Or am I missing something?

punker76: The new behavior (ToolTipBehavior) is only for the ToolTip. It doesn't handle the visibility or command usage of the element. So now if the `IsInstallAllowed` is set to false, you will see the disabled feature tooltip, but the button is still available and the user can install the package if it's not installed.

```
<Button Padding="10" Margin="5 0" Command="{commands:DataContextCommandAdapter Install}">

</Button>
```

gep13:
@punker76 I have looked at what you suggested, and I have updated the relevant sections.  I have tested it out this morning, and this is what is now happening.

While running as a normal OSS user of Chocolatey GUI, you will get the following:
 
image

image

image

i.e. all buttons enabled, and tooltips correctly showing what should be done.

But when running with the Chocolatey GUI Licensed Extension, which will respect the current configuration in the `backgroundServiceAllowedCommands` setting, you will get the following:

image

image

i.e. the options that are not allowed will be disabled, the the tooltip will provide information about what is going on.

What do you think?

punker76 (approved):
LGTM

PR #822, opened 2021-03-08
Fix logic of IsShowOnlyPackagesWithUpdateEnabled when PreventAutomatedOutdatedPackages is not set

### LINKED ISSUE #821: IsShowOnlyPackagesWithUpdateEnabled doesn't work correctly when PreventAutomatedOutdatedPackages is true
Just found this issue after merged dark mode PR where I changed the logic for this too. I'll provide a PR to fix this.

punker76:
The IsShowOnlyPackagesWithUpdateEnabled will be false if the PreventAutomatedOutdatedPackagesCheck is not set. So you will not see any updated packages without forcing the search for it.

This PR fix this issue.

Closes #821


gep13 (approved):
LGTM!

gep13:
@punker76 thanks for finding and fixing this!

PR #823, opened 2021-03-09
(#545) Add advanced installation dialog

### LINKED ISSUE #545: Allow for passing in package parameters / installer arguments  (maybe other arguments directly to choco)
It would be great to be able to pass through package parameters and installer arguments as part of installation. There are some folks out there who enjoy the UI but still want some of the customization of installed packages. 

Other things to consider:

* Setting to force which architecture of a package is installed (x64, x86)
* Ability to select which version of a package is installed

Based on discussion during a Twitch session on 20th April 2020, the following things were discussed/mentioned regarding this topic (@AdmiringWorm @mkevenaar @mwallner @punker76)

- [x]  Selected version should be the latest version available by default
- [ ]  In order for this to work, we need a new release of Chocolatey.Lib
- [x]  We should restrict the number of available package versions, perhaps 20 by default (but this should be configurable), with an option to load more results
- [ ]  There should be a `More ..` option, or similar, next to the combobox for loading more package versions
- [x]  Versions should be sorted from newest to oldest
- [ ]  Perhaps provide the ability to allow a package version number to be entered directly into the box, but this should be validated before usage
- [ ]  Would be could to have autocomplete/filtering on the combobox to make it easy to find a package version
- [x]  There should be a checkbox to control whether pre-release package versions are included in the combobox
- [x]  Would be nice to fetch the available package versions once the Advanced dialog is opened, with a loading indicator

In addition, we went through the available options/switches for the install command, and decided that the following should be available:

* Command Timeout
* Log File
* Version
* Pre-Release (could be set automatically, based on selected package version)
* Force x86
* Install Args
* Override Args
* Not Silent (we were undecided on this one)
* Package Parameters
* Args Global
* Params Global
* Allow Downgrade (could perhaps infer this, based on current state of system)
* SxS (we we undecided on this one)
* Ignore dependencies
* Force dependencies
* Skip PowerShell
* Checksums, and all related options - discussion to be had around this

Think that was everything that was discussed, but can add more

gep13:
Which will allow controlling the parameters/options, that are passed into the install command.

Fixes #545


punker76:
We need some help for the UI of this advanced install dialog.

Current implementation:

2021-03-15_12h28_46

So here are 3 concepts / mockups / suggestions from me:

1) Current dialog with updated UI to set the options

2021-03-15_11h47_32

2) New UI with a different dialog

2021-03-15_12h01_14

3) New UI as Flyout

2021-03-15_12h17_19

What's your favorite?

mkevenaar:
My vote goes to no 2

timunie:
I vote for #2

MarcoSpoenemann:
The 2nd one looks nice.

mwallner:
> What's your favorite?

Amazing work! - I'd vote for 2

AdmiringWorm:
I guess I would need to be different then.

I personally like the third option with it available as a flyout, although the second option isn't bad either.

Just one thing I feel with all of the options, the margin/padding looks like it is set a bit low and makes the text too close to the edges.

punker76:
**Update**

chocolateygui_childwindow

gep13:
@punker76 just taken this for a spin, and this looks really good to me. Are we at a point where we can get this merged in? Or do you still have changes that you would like to make? Should I move this out of draft status?

punker76:
@gep13 I'm happy with it so far and done. Maybe there will be some changes in the individual properties, but we can also change that later.

gep13:
@punker76 sounds good. Going to move forward with getting this merged in!  Thanks for all your help in getting this to where it is!

PR #839, opened 2021-04-09
(#786) Add confirmation for destructive operations

### LINKED ISSUE #786: Ask to confirm Uninstall/Reinstall from Right-Click Menu
When you open the right-click menu for any installed application, you should get a confirmation prompt if you click on Uninstall or Reinstall. If they are clicked accidentally when the user actually wanted to click on Pin or Details, the user might lose important data.

gep13:
A confirmation dialog is being added to the reinstall and uninstall
operations since these can be harmful operations to the system.

Fixes #786

### DIFF
--- a/Source/ChocolateyGui.Common.Windows/Services/IProgressService.cs
+++ b/Source/ChocolateyGui.Common.Windows/Services/IProgressService.cs
@@ -27,6 +27,8 @@ namespace ChocolateyGui.Common.Windows.Services
 
         Task<MessageDialogResult> ShowMessageAsync(string title, string message);
 
+        Task<MessageDialogResult> ShowConfirmationMessageAsync(string title, string message);
+
         Task StartLoading(string title = null, bool isCancelable = false);
 
         Task StopLoading();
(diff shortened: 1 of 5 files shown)


gep13:
@punker76 can you offer any help here?

When I try to run the application, I get the following error displayed:

image

I have tried to add the participation element, in the same way that I did here:

https://github.com/chocolatey/ChocolateyGUI/commit/59ddbf71db1ecaaebca24ccba3557396f02085cc

But no matter where I put it, it results in the same thing each time. 

Thanks

punker76:
@gep13 I implemented as new method ShowConfirmationMessageAsync on IProgressService  and use this new method to show a confirmation dialog e.g. when the user choose Reinstall or Uninstall. Using the DialogCoordinator is not possible for this, because the PackageViewModel can not be registered to the same view as for the local or remote source view.

gep13:
@punker76 thank you very much for fixing this up!

opusforlife2:
Thank you! This is a major lifesaver!

PR #840, opened 2021-04-09
(#838) Add ability to filter features/settings

### LINKED ISSUE #838: Provide ability to search/filter features and settings
Due to the growing number of features and settings within Chocolatey GUI:

image

I believe that it would be a good idea to provide the ability to search/filter the list of items so that you can quickly find the item that you are looking for.  This would search/filter on the title of the feature/setting.

We already do this on the local source view for packages, so it would be very similar to this:

image

gep13:
For both Chocolatey GUI and Chocolatey. This will make it much easier
to find the setting that you are interested in changing. This has been
implemented by using an ICollectionView for each collection, and then
applying a filter method to each one, based on what is typed into the
search textbox.

Fixes #838

### DIFF
--- a/Source/ChocolateyGui.Common.Windows/ViewModels/SettingsViewModel.cs
+++ b/Source/ChocolateyGui.Common.Windows/ViewModels/SettingsViewModel.cs
@@ -7,11 +7,13 @@
 
 using System;
 using System.Collections.ObjectModel;
+using System.ComponentModel;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
+using System.Windows.Data;
 using Caliburn.Micro;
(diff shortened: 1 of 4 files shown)


gep13:
filter-chocolatey-gui-settings

gep13:
@punker76 this obviously needs some work on the UI side of things, but I wanted to get your opinion on what I have done here.  Do you think this makes sense?

mkevenaar:
My 2 cents: Why not use 1 search field to search in both panes and position it on the same spot as the "normal" search field

punker76:
@gep13 My possible small contribution to this:

2021-04-12_09h34_15

2021-04-12_09h49_18

gep13:
@punker76 that looks really good to me!  Would you be able to add this as a commit to this PR?

Having them as separate search boxes makes sense to me still.

gep13:
@punker76 thank you very much for putting the UI touches on this, really appreciate it!

PR #637, opened 2019-01-16
[WIP] (GH-533) Implemented ability to change language dynamically

AdmiringWorm:
This pull request adds support for changing between languages dynamically during runtime.

This is still very much a WIP, and some features currently do not work.
Such as:
- Loading the correct language during startup
- Translation of 'This Computer'
- Grabbing all supported languages
- Validation

PR #844, opened 2021-04-19
(#533) Ability to change language dynamically

### LINKED ISSUE #533: Allow users to change the locale to use
One thing I feel is missing is to allow the user to change the locale to display Chocolatey GUI in (I like to change from Norwegian to English and back, depending on the circumstance).

It is possible to allow such an option by doing a naïve check during runtime by checking which locale folder exists: see https://stackoverflow.com/a/32963564 for more info there.

Or creating a static configuration list of the available locales during build, which can be done using the Cake.Transifex addin by using the `TransifexStatus` alias (If I have remembered to correctly implement it, I think I have).

Optimally changing the locale would be dynamic, but that would require quite a few changes, including a a major change in the code base, which probably shouldn't be a priority at this time.
Alternatively, it would notify the user to restart Chocolatey GUI to change the locale.

gep13:
This is based solely on the work done by AdmiringWorm in this PR:

https://github.com/chocolatey/ChocolateyGUI/pull/637

And has been re-created here due to fundamental changes in the project
since that PR was created.

Fixes #533


gep13:
@AdmiringWorm @punker76 I have re-created the PR that was submitted here:

https://github.com/chocolatey/ChocolateyGUI/pull/637

Due to the fact that resolving all the merge conflicts in that PR, given the changes in the project since it was created was doing to be too difficult.

Going forward, we can use this PR for any additional work/changes.

gep13:
@AdmiringWorm would you be able to get this one updated and rebased after the recent merges into develop?

AdmiringWorm:
@gep13 I have rebased the PR and updated the strings that had been added in merged commits that weren't using the code with the ability to dynamically change the language.

Source/ChocolateyGui.Common/Properties/Resources.resx
```diff
@@ -1245,6 +1245,92 @@ Please contact your System Administrator to enable this operation.</value>
     <value>All the currrently installed packages have been exported to {0}</value>
     <comment>{0} = The path to where the exported packages have been placed</comment>
   </data>
+  <data name="SettingsView_Language" xml:space="preserve">
+    <value>Language</value>
+  </data>
+  <data name="RemoteSourceView_TooltipGoBackAPage" xml:space="preserve">
+    <value>Got to the Previous Page</value>
```
gep13, line 1252:
```suggestion
    <value>Go to the Previous Page</value>
```

Source/ChocolateyGuiCli/Runner.cs
```diff
@@ -56,6 +58,10 @@ public static void Run(string[] args)
                     commandArgs.Add(arg);
                 }
 
+                // We need to initialize the current culture before using we
```
gep13, line 61:
```suggestion
                // We need to initialize the current culture before we

```

Source/ChocolateyGui.Common.Windows/Services/DialogService.cs
```diff
@@ -200,5 +201,15 @@ public async Task<LoginDialogData> ShowLoginAsync(string title, string message,
...
+
+        private static string L(string key)
+        {
+            return TranslationSource.Instance[key];
+        }
+
+        private static string L(string key, params object[] parameters)
+        {
+            return TranslationSource.Instance[key, parameters];
+        }
```
gep13, line 213: This is repeated a few times for services... Is it worth creating a BaseService class that has these methods?  Similar to what was done with the new ViewModelScreen class?

gep13 (commented):
Just a couple things, otherwise this looks really good!

Source/ChocolateyGui.Common.Windows/Services/DialogService.cs
AdmiringWorm: I was honestly thinking about it but didn't want to pollute a base class with just static methods.

I could still do it though, but it wouldn't be the same as the ViewModelScreen base class as that also support passing in a distinct TranslationSource which wouldn't be possible for the classes in the service that require a static method.

vexx32: If they're all static method, we could stick them in a shared helper class, it needn't (necessarily) be a base class.

AdmiringWorm: @vexx32 technically the helper class is the `TranslationSource` class, the methods here are just for convenience.

gep13: @AdmiringWorm gotcha. Makes sense. Based on what you have said here, I am happy to leave this the way it is.

AdmiringWorm:
@gep13 all should be updated now, including the typo for the purge command that we spoke about

gep13 (commented):
LGTM!

gep13:
@AdmiringWorm thanks again for working on this, this looks AMAZING, and it will be great to get this shipped in the next release of Chocolatey GUI!

image

PR #847, opened 2021-04-20
(#842) added a check for NET Framework 4.8

### LINKED ISSUE #842: Ensure WiX installer verifies that .NET Framework 4.8 is installed
As a follow up to this issue: https://github.com/chocolatey/ChocolateyGUI/issues/841 we need to update the WiX installer to ensure that .NET Framework 4.8 is installed.

This is really belts and braces, since we have updated the dependency on the Chocolatey package to ensure that it is installed, however, if installing the MSI directly, it wouldn't install correctly if this check isn't there.

When I attempted to make this change, the result was the following:

image

I installed the the 3.14 version from the development downloads section: https://wixtoolset.org/releases/development/ as it was mentioned here: https://github.com/wixtoolset/wix3/blob/develop/src/chm/documents/customactions/wixnetfxextension.html.md#properties that this was required, but it still chose not to work.

nils-a:
in the installer.

While the code I used to implement the check is part of multiple comments/discussions I decided to copy from the "original", so I added an extra comment for the license of the original (MS-RL).
The whole file `NetFx48.wsx` can be removed, once the Wix version is bumped to one greater than v3.14.0.4118.

fixes #842


gep13 (approved):
LGTM!

gep13:
@nils-a thank you VERY much for getting this added, really appreciate it!

PR #876, opened 2021-08-23
(#875) Don't use ConfigureAwait(false) inside a progress dialog context

### LINKED ISSUE #875: All operations on packages with a progress dialog are broken
Currently all operations on a package like Pin, Unpin, install... etc are broken. The operations will be execute but the progress dialog will not close.

2021-08-23_09h50_36

This is an issue with the used `ConfigureAwait(false)` method on each operation inside the progress dialog context.

https://github.com/chocolatey/ChocolateyGUI/blob/0256e2f694ec9cabc1fd5c1dc7199818b6023939/Source/ChocolateyGui.Common.Windows/ViewModels/Items/PackageViewModel.cs#L582

punker76:
This PR fixes the broken progress dialog on all package operations.

Closes #875


AdmiringWorm:

PR #878, opened 2021-08-26
Bad type file

Starmania:
Editing in chocolateyInstall.ps1 fileType var from 'msi' to 'exe'.


Starmania:
My bad, but when I try to install, he download me an exe file and create an error when executing... I don't know why...

gep13:
Can you show the output from the command that you ran?

Chocolatey GUI has some dependencies which may result in an exe being downloaded, but Chocolatey GUI itself ships as an msi.

PR #910, opened 2022-01-28
(#545) Update to better support advanced installation

### LINKED ISSUE #545: Allow for passing in package parameters / installer arguments  (maybe other arguments directly to choco)

AdmiringWorm:
## Description Of Changes

This pull request updates the advanced installation view to better support those scenarios, by fixing a few bugs as well as adding some very basic disabling/hiding of arguments when they do not make sense to be used.

Additionally, there is a couple of bug fixes for the changing of the Languages that are used to prevent sources to be reloaded during language change, and to set the correct WPF Element languages.

## Motivation and Context

This is done to better support scenarios where advanced installation is to be done, and to make it a bit easier for the user to know what can/should not be used together.

## Testing

1. Tested by installing different packages, with a mix of different features enabled and disabled.
2. Tested by toggling features on and off to see that the expected elements was either disabled, or removed from the view.
3. Tested that settings page could not be accessed when advanced installation window is open (and about section).
4. Tested that changing a language when the last viewed page is the remote source did not trigger a reload of the packages.
5. Tested that a cache directory can be set, and browsed to using the correct button.
6. Tested that that changing the language would also correctly set the language (seen by navigating to a package details page after the change to see if the numbers and datetime is in the correct format).
7. Probably something else as well, but can not remember.

NOTE: While the correct flow direction is also set, consider this unsupported still until there is feedback from users using a language with right to left support to figure out if it is correctly displayed or not.

## Change Types Made

* [x] Bug fix (non-breaking change)
* [x] Feature / Enhancement (non-breaking change)

## Related Issue

Fixes #545
A couple of bug fixes fo #533

## Change Checklist

* [x] Requires a change to the documentation
* [ ] All new and existing tests passed. (There aren't any)
* [ ] PowerShell v2 compatibility checked. (No PowerShell scripts have been touched)


Source/ChocolateyGui.Common.Windows/Views/SettingsView.xaml
```diff
@@ -64,7 +64,8 @@
 
                     <TextBlock Grid.Column="0" Grid.Row="0" Style="{StaticResource BaseAltTextBlockStyle}" Text="{Binding Name}" TextWrapping="Wrap" />
                     <metro:ToggleSwitch Grid.Column="1" Grid.Row="0" VerticalAlignment="Center" Margin="4" IsOn="{Binding Enabled}" cal:Message.Attach="[Event Toggled] = [Action ChocolateyFeatureToggled($dataContext)]"/>
-                    <TextBlock Grid.Column="0" Grid.ColumnSpan="2" Grid.Row="1" Style="{StaticResource BodyTextBlockStyle}" Text="{Binding Description}" TextWrapping="Wrap" />
+                    <!-- Remove the following FlowDirection once we can translate Chocolatey feature descriptions -->
+                    <TextBlock Grid.Column="0" Grid.ColumnSpan="2" Grid.Row="1" Style="{StaticResource BodyTextBlockStyle}" Text="{Binding Description}" TextWrapping="Wrap" FlowDirection="LeftToRight" />
```
gep13, line 68: Can you capture an issue to make sure that this doesn't get forgotten about?

gep13 (changes requested):
While this isn't directly associated with the work that you have done here, there is one thing missing in the new advanced install function, and I think we should address it here in this PR...

Currently, within the Package View, the ability to click the Install command is controlled by this behaviour:

https://github.com/chocolatey/chocolateygui/blob/develop/Source/ChocolateyGui.Common.Windows/Views/PackageView.xaml#L86-L89

We should have the same behaviour on the install advanced command.  If the user isn't allowed to install a package, they shouldn't be allowed to do an advanced install of a package.  Can you get this added to this PR?

gep13:
Other than the above comment, this looks really good, and once fixed up, we can get this approved and merged.

Source/ChocolateyGui.Common.Windows/Views/SettingsView.xaml
AdmiringWorm: Done https://github.com/chocolatey/ChocolateyGUI/issues/912

AdmiringWorm:
@gep13 the PR have been updated with your requested changes

gep13 (approved):

gep13:
@AdmiringWorm thanks for getting this updated, and for verifying that this is all working as expected!

PR #926, opened 2022-03-07
(build) Adds TeamCity build definition

gep13:
## Description Of Changes

This commit adds a build definition to be used by the Chocolatey
internal TeamCity server.

This calls the build.official.bat script, as it will have access to the
required files in order to create an official signed version of
Chocolatey GUI.

Setup of this build uses the build of Chocolatey GUI Licensed Extension
as a template, and then modified the necessary portions to make it
specific to Chocolatey GUI.

## Motivation and Context

All Chocolatey builds are going to happen on an internal TeamCity server, to allow for full signing of the generated assemblies.

## Testing

The TeamCity build configuration has been tested on a previous TeamCity server, and brought over in the settings.kts file.

## Change Types Made

* [ ] Bug fix (non-breaking change)
* [ ] Feature / Enhancement (non-breaking change)

## Related Issue

N/A

## Change Checklist

* [ ] Requires a change to the documentation
* [x] All new and existing tests passed.
* [ ] PowerShell v2 compatibility checked.

### DIFF
--- a/.teamcity/settings.kts
+++ b/.teamcity/settings.kts
@@ -1,0 +1,88 @@
+import jetbrains.buildServer.configs.kotlin.v2019_2.*
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildSteps.script
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildSteps.powerShell
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildFeatures.xmlReport
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildFeatures.XmlReport
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildFeatures.commitStatusPublisher
+import jetbrains.buildServer.configs.kotlin.v2019_2.buildSteps.nuGetPublish
+import jetbrains.buildServer.configs.kotlin.v2019_2.triggers.vcs
+import jetbrains.buildServer.configs.kotlin.v2019_2.vcs.GitVcsRoot
+
+version = "2021.2"
+
+project {
+    buildType(ChocolateyGUI)
+}
+
+object ChocolateyGUI : BuildType({
+    id = AbsoluteId("ChocolateyGUI")
+    name = "Build"
+
+    artifactRules = """
+        code_drop/MsBuild.log
+        code_drop/MSBuild.msi.log
+        code_drop/ChocolateyGUI.msi
+        code_drop/TestResults/issues-report.html
+        code_drop/Packages/**/*.nupkg
+    """.trimIndent()
+
+    params {
+        param("env.vcsroot.branch", "%vcsroot.branch%")
+        param("env.Git_Branch", "%teamcity.build.vcs.branch.ChocolateyGUI_ChocolateyGuiVcsRoot%")
+        param("teamcity.git.fetchAllHeads", "true")
+        password("env.TRANSIFEX_API_TOKEN", "credentialsJSON:c81283e6-cf59-5c9e-9766-6f465018a295", display = ParameterDisplay.HIDDEN, readOnly = true)
+        password("env.GITHUB_PAT", "%system.GitHubPAT%", display = ParameterDisplay.HIDDEN, readOnly = true)
+    }
+
+    vcs {
+        root(DslContext.settingsRoot)
+
+        branchFilter = """
+            +:*
+        """.trimIndent()
+    }
+
+    steps {
+        powerShell {
+            name = "Prerequisites"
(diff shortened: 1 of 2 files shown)


.teamcity/settings.kts
```diff
@@ -0,0 +1,88 @@
...
+        code_drop/TestResults/issues-report.html
+        code_drop/Packages/**/*.nupkg
+    """.trimIndent()
+
+    params {
+        param("env.vcsroot.branch", "%vcsroot.branch%")
+        param("env.Git_Branch", "%teamcity.build.vcs.branch.ChocolateyGUI_ChocolateyGuiVcsRoot%")
+        param("teamcity.git.fetchAllHeads", "true")
+        password("env.TRANSIFEX_API_TOKEN", "credentialsJSON:PUT_SOMETHING_HERE", display = ParameterDisplay.HIDDEN, readOnly = true)
+        password("env.GITHUB_PAT", "credentialsJSON:PUT_SOMETHING_HERE", display = ParameterDisplay.HIDDEN, readOnly = true)
```
JPRuskin, line 34:
```suggestion
        password("env.GITHUB_PAT", "credentialsJSON:77847f9b-fb3c-59cb-be70-a708b3d3d1ed", display = ParameterDisplay.HIDDEN, readOnly = true)
```

I added the function to figure this out [here](https://gist.github.com/JPRuskin/8992e5665a3a73316bba3089de074078).

.teamcity/settings.kts
```diff
@@ -0,0 +1,88 @@
...
+            branchFilter = ""
+        }
+    }
+
+    features {
+        pullRequests {
+            vcsRootExtId = "${DslContext.settingsRoot}"
+            provider = github {
+                authType = token {
+                    token = "PUT_SOMETHING_HERE"
```
JPRuskin, line 83: Presumably this needs to be referencing the parameter above?

```suggestion
                    token = "%env.GITHUB_PAT%"
```

gep13: Makes sense to me.

.teamcity/settings.kts
```diff
@@ -0,0 +1,88 @@
...
+            branchFilter = ""
+        }
+    }
+
+    features {
+        pullRequests {
+            vcsRootExtId = "${DslContext.settingsRoot}"
+            provider = github {
+                authType = token {
+                    token = "%env.GITHUB_PAT%"
```
JPRuskin, line 83:
```suggestion
                    token = "%system.GitHubPAT%"
```

Windos: Sorry, I'm jumping into this without contex of any discussions that've happened outside of the PR.

It seems like the intention here was to use the `env.GITHUB_PAT` parameter set above. If it's not being used down here, is the parameter still needed above (i.e. is it used as an environment variable in the build)?

JPRuskin: You're bang on - Gary mentioned that I should specify the variable name to use, and he'd remove the above.

.teamcity/settings.kts
```diff
@@ -0,0 +1,88 @@
...
+        code_drop/TestResults/issues-report.html
+        code_drop/Packages/**/*.nupkg
+    """.trimIndent()
+
+    params {
+        param("env.vcsroot.branch", "%vcsroot.branch%")
+        param("env.Git_Branch", "%teamcity.build.vcs.branch.ChocolateyGUI_ChocolateyGuiVcsRoot%")
+        param("teamcity.git.fetchAllHeads", "true")
+        password("env.TRANSIFEX_API_TOKEN", "credentialsJSON:c81283e6-cf59-5c9e-9766-6f465018a295", display = ParameterDisplay.HIDDEN, readOnly = true)
+        password("env.GITHUB_PAT", "credentialsJSON:77847f9b-fb3c-59cb-be70-a708b3d3d1ed", display = ParameterDisplay.HIDDEN, readOnly = true)
```
JPRuskin, line 34:
```suggestion
```

gep13:
@JPRuskin I have updated the PR based on one of the suggestions that was made, however, there is another question that I have left on the other suggestion.  If you can let me know your thoughts on that one, it would be appreciated.  Happy to jump on a call if you wanted to talk through it.

.teamcity/settings.kts
JPRuskin: That's alright- I think / suspect what we can do here, then (though will need to test this) that you should be able to set a parameter `env.GitLab_PAT` to the value of `%system.GitLabPAT%`. Will that work?

gep13: Ok, cool, let's take this for a spin! :shipit:

PR #955, opened 2022-08-30
Refresh sources from file

corbob:
TODO: fill out

## Change Types Made

* [ ] Bug fix (non-breaking change)
* [ ] Feature / Enhancement (non-breaking change)

## Related Issue

Fixes #

## Change Checklist

* [ ] Requires a change to the documentation
* [ ] All new and existing tests passed.
* [ ] PowerShell v2 compatibility checked.


Source/ChocolateyGui.Common.Windows/Services/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,53 @@
...
+                Path = _fileSystem.get_directory_name(_configFile),
+                Filter = _fileSystem.get_file_name(_configFile),
+                NotifyFilter = NotifyFilters.LastWrite | NotifyFilters.FileName,
+            };
+            _fileSystemWatcher.Changed += ConfigFileChanged;
+            _fileSystemWatcher.EnableRaisingEvents = true;
+            _lastKnownConfigFileHash = _xmlService.deserialize<ConfigFileSettings>(_configFile).GetHashCode();
+        }
+
+        private void ConfigFileChanged(object sender, FileSystemEventArgs e)
```
gep13, line 43: Is there a reason that this is private, and why `ConfigFileChanged` isn't added to the `IConfigFileWatcher` interface?

Source/ChocolateyGui.Common.Windows/Services/ConfigFileWatcher.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace ChocolateyGui.Common.Windows.Services
+{
+    public class ConfigFileWatcher : IConfigFileWatcher
+    {
+        private readonly IFileSystem _fileSystem;
+        private readonly IEventAggregator _eventAggregator;
+        private readonly FileSystemWatcher _fileSystemWatcher;
+        private readonly IXmlService _xmlService;
+        private readonly string _configFile = chocolatey.infrastructure.app.ApplicationParameters.GlobalConfigFileLocation;
```
gep13, line 24: This might be better as `_configFilePath`?

Source/ChocolateyGui.Common.Windows/Bootstrapper.cs
```diff
@@ -121,6 +121,7 @@ protected override async void OnStartup(object sender, StartupEventArgs e)
 
                 var packageService = Container.Resolve<IChocolateyService>();
                 var features = await packageService.GetFeatures();
+                var fileSystemWatcher = Container.Resolve<IConfigFileWatcher>();
```
gep13, line 124: This is being resolved, but isn't being used, or injected anywhere, it that intentional?

Source/ChocolateyGui.Common.Windows/Services/ChocolateyService.cs
```diff
@@ -497,8 +495,8 @@ public async Task<ChocolateySource[]> GetSources()
             // we need to read all information from the config file, i.e. the username and password
             var config = await GetConfigFile();
             var allSources = config.Sources.Select(_mapper.Map<ChocolateySource>).ToArray();
-
-            var filteredSourceIds = _configSettingsService.source_list(_choco.GetConfiguration()).Select(s => s.Id).ToArray();
+            var configSettingsService = new ChocolateyConfigSettingsService(_xmlService);
```
gep13, line 498: What was the reason that this is being done, rather than relying on the injected version of the service?

gep13 (changes requested):

Source/ChocolateyGui.Common.Windows/Services/ConfigFileWatcher.cs
corbob: I think because the code I was looking at in Chocolatey Agent had used a private method for the file changed handler that I just kept it private. Given the IConfigFileWatcher, it probably makes sense to make it public and add it to the interface.

Source/ChocolateyGui.Common.Windows/Services/ConfigFileWatcher.cs
corbob: Naming is hard, but yes that is a much better name.

Source/ChocolateyGui.Common.Windows/Bootstrapper.cs
corbob: In my testing, it seemed that the watcher wasn't being instantiated until it was resolved, and it seemed to make sense to resolve it where all of the other services were being resolved.

Would it be better to resolve it when we add it to the container?

gep13: We spoke about this, and it's all good.

corbob:
After reverting commit 583b1b9f160a92df7b1ea9d39ddc46bf28a7949b I performed the tests in the PR. My observations are:

1. With Just Chocolatey CLI installed, all existing sources in the file when Chocolatey GUI is launched successfully maintain state. If you enable or disable them outside of Chocolatey GUI, they are updated inside of Chocolatey GUI. However, if you add a new source, it fails to be detected.
2. With Chocolatey Licensed Extension installed, the tests behaved the same as Chocolatey CLI. The Licensed specific tests however did not work as desired. The only change that was correctly picked up is enable, disable, or removal of an existing source. Changes to the Admin Only or Self Serve status does not get reflected.

Issue #697, opened 2019-10-02
Chocolatey GUI doesn't write to chocolatey.log


jacktose:
Chocolatey CLI writes to `C:\ProgramData\chocolatey\logs\chocolatey.log`, but Chocolatey GUI does not.

This is especially confusing because the CLI output that the GUI shows says to look in the chocolatey log: ![image](https://user-images.githubusercontent.com/12414794/66088304-7e7f3400-e52f-11e9-862f-81e30a09c9c2.png)

@gep13 says it's a known issue:
> This is a side effect of the way that Chocolatey GUI calls into Chocolatey. We pass in an additional logger which should mean that logging happens to both the chocolatey log and the chocolatey gui log.

Possibly related to #695.

ferventcoder:
https://github.com/chocolatey/choco/issues/2122

PR #1129, opened 2026-04-20
Fix duplicated messages in console output

### LINKED ISSUE #933: Multiple duplicate lines in choco console UI (but not log file)
### What You Are Seeing?

When installing or updating a package, the console often shows multiple (2 or 3) duplicates of the log messages that show after the `Exporting function`, `Importing function`, and `Importing alias` lines.

One example:
```
[INFO ] - VERBOSE: Refreshing environment variables from the registry.
[INFO ] - VERBOSE: Refreshing environment variables from the registry.
[INFO ] - VERBOSE: Refreshing environment variables from the registry.
[INFO ] - VERBOSE: Refreshing environment variables from the registry.
[INFO ] - Restricting write permissions to Administrators
[INFO ] - Restricting write permissions to Administrators
[INFO ] - Restricting write permissions to Administrators
[INFO ] - We are setting up the Chocolatey package repository.
```

### What is Expected?

Each line should only show up once.

### How Did You Get This To Happen? (Steps to Reproduce)

1. Turn on "Show console output".
2. Install or update any package.
3. Notice the ongoing log, especially around the lines related to downloading and installing.

### System Details

* OS Build (In PowerShell run `[System.Environment]::OSVersion.version.tostring()`): Windows 7
* Windows PowerShell version (Run: `$PSVersionTable`): Chocolatey inbuilt
* Chocolatey version (Run `choco --version`): 

### Output Log

N/A - The log file correctly shows only one line for each message. This seems to be a UI problem only.

danrhodes:
**Description Of Changes**
Fixed a long-standing bug where every line of console output was duplicated in the GUI console panel during package install, update, and uninstall operations.

image
image

The fix is a one-word change (true → false) on the addToExistingLoggers parameter in three SetCustomLogging calls within ChocolateyService.cs.

**Motivation and Context**
Lets.GetChocolatey() returns a shared singleton. The ChocolateyService constructor registers a SerilogLogger against this singleton with addToExistingLoggers: true. Each of the InstallPackage, UninstallPackage, and UpdatePackage methods then calls Lets.GetChocolatey() again — returning the same shared instance — and registers a second SerilogLogger against it, also with addToExistingLoggers: true.

Because addToExistingLoggers: true appends to the internal logger list rather than replacing it, by the time any operation executes there are two active SerilogLogger instances both pointing at the same _progressService. Chocolatey dispatches every log message to all registered loggers, so each message is written to the UI console twice.

The root cause went undetected for years because the two logger registrations are in separate methods and each looks reasonable in isolation. The duplication is a visual artefact rather than a crash or data corruption, so it never surfaced as an automated test failure.

**Testing**
- Installed a package via the GUI and confirmed each console output line appears exactly once.
- Updated a package via the GUI and confirmed each console output line appears exactly once.
- Uninstalled a package via the GUI and confirmed each console output line appears exactly once.
- Confirmed all existing tests pass.
image

image

**Operating Systems Testing**
Windows 11 Pro

**Change Types Made**
Bug fix (non-breaking change).

Related Issue
Fixes #933


danrhodes:
CLA signed 👍

danrhodes:
I'm going to check over this again later as I see a comment from @corbob re #697 and logging breaking with this fix he tried previously that should be addressed.

danrhodes:
@corbob @gep13 - After further investigation, the initial approach of passing **addToExistingLoggers: false** to SetCustomLogging was found to be incorrect.

When **false**, Chocolatey builds an AggregateLog containing only the GUI's custom logger and overwrites the static Log._logger field — silently dropping the Chocolatey file logger and causing operations to stop writing to chocolatey.log, reintroducing the regression fixed in #697.

The correct fix is to stop calling Lets.GetChocolatey() inside InstallPackage, UninstallPackage, and UpdatePackage altogether.

Since Lets.GetChocolatey() returns a shared singleton, each call was mutating the same instance the constructor had already configured, appending a second logger on top of the first.

By reusing _choco directly — which is configured once in the constructor with a single SerilogLogger — the logger list is never modified again, and duplication cannot occur.

The local SerilogLogger variable is retained in each method solely for RunCommand's error-interception mechanism, which is separate from the Chocolatey logger infrastructure.

danrhodes:
Any update on this please?

corbob:
> Any update on this please?

@danrhodes, thank you for taking the time to submit this PR. I unfortunately haven't had an opportunity to review it or test it yet. As I recall when I looked at this issue, there was more than just the semi-singleton nature of `Lets.GetChocolatey` at play. I've been trying to remember what tests I had done that found it not feasible, and I'm not entirely sure. But at the very least the below tests will need to be done to help us have confidence in the results. For each of these tests, we need to ensure that the logging output is present in all locations (the Console in Chocolatey GUI, the Chocolatey log at `c:\programdata\chocolatey\logs\chocolatey.log`, and the Chocolatey GUI logs at `c:\programdata\Chocolatey GUI\Logs\`):

- Install a package
- Install a different package
- Upgrade a package
- uninstall a package
- check for outdated packages
- Install a package with customized parameters

danrhodes:
Thanks for replying, I will.do the needfull tonight and post the results.

vexx32:
@danrhodes I haven't looked at the code here yet, I'll try to give it a look tomorrow, but you can look at the following branches for my previous investigations which. I do think solved the issue, ultimately, when used together, but I haven't had time to thoroughly validate or wrap tests around them as yet:

https://github.com/vexx32/choco/tree/logging-fixes

and the GUI branch:

https://github.com/vexx32/ChocolateyGUI/tree/cle646-investigation

TL;DR: choco's base logging API is kind of. not working consistently or correctly in some cases.

I... think a decent portion of the blocker on fixing this *probably* was around wanting to make sure that the new logging API I wanted to add have some validation and testing around them, but I've not been sure how to sort those out / haven't had time available to look at it in detail.

EDIT: Skimmed the code quickly and I *think* while this is a partial fix there were still scenarios I ran into when looking into this with similar ideas where it would still break.

